PR #241, opened 2017-01-31
Adaptive nepochs fix

j3mdamas:
@stefdoerr can you check if this is fine?

Before this change, at the first updateperiod after the submission of the last epoch, the adaptive would end regardless if there were still simulations running or not (depending on the nmin and nmax, it could mean 1 or 2 epochs less of trajectories).

I've tested it and it seems like it does not break, can you check and accept the pull request?


j3mdamas:
maybe wait for Travis and then accept

stefdoerr:
Yes, yes I will check. But tomorrow. I'm finishing OPM today

j3mdamas:
sure, whenever you have time

stefdoerr:
I checked anyways :) It's good.

My only miiiinor gripe is the time wasted checking inprogress() everytime we call it once it's done but it's unavoidable unless we set somewhere a `finished` flag in a file or something which would be too much mess.

Thanks for the fix :)

j3mdamas:
You didn't merge, just closed, but I'll merge now

PR #266, opened 2017-03-08
Correct sigma computation in voxelDescriptors

hawk31:


htmd/molecule/voxeldescriptors.py
```diff
@@ -149,34 +149,41 @@ def _findDonors(mol, bonds):
 
 
 def _getSigmas(mol):
-    """ Returns the VDW radii of each atom in the molecule
+    # Sigmas taken from VMD vdW radii.
```
stefdoerr: Add a link in the comment to where you got them from

htmd/molecule/voxeldescriptors.py
```diff
@@ -149,34 +149,41 @@ def _findDonors(mol, bonds):
...
+     'N':   1.55,
+     'NA':  1.55,
+     'OA':  1.52,
+     'P':   1.8,
+     'S':   1.8,
+     'SA':  1.8,
+     'Zn':  1.39,
+     'Se':  1.9}
 
-    Parameters
```
stefdoerr: Why did you remove the docs?

htmd/molecule/voxeldescriptors.py
j3mdamas: I don't know if we are going for chemical correctness here, but lennard-jones sigmas and van der waals radius are not the same thing

hawk31:
Okay, I'll make this pull request more complete tomorrow. You can close it for the moment.

PR #278, opened 2017-03-15
Added Radii.

miha-skalic:


stefdoerr:
Please document where the values came from and what those atoms are

hawk31:
Let's try and make it consistent. Two decimals, rounding to the nearest decimal instead of removing them, etc.

j3mdamas:
+1 @stefdoerr comment.

miha-skalic:
Updated.
@stefdoerr @hawk31

PR #279, opened 2017-03-15
Fix to occupancy calculation.

miha-skalic:


htmd/molecule/voxeldescriptors.py
```diff
@@ -4,12 +4,13 @@
 # No redistribution in whole or part
 #
 from htmd.molecule.util import boundingBox
-import numpy as np
```
stefdoerr: Never trust pycharm on unused dependencies. This file will crash. The last function uses `np.`

miha-skalic:
@stefdoerr Updated

stefdoerr:
Cool :)

PR #355, opened 2017-06-01
Return resid collisioning

cotufalaguera:
I have changes the following function and class methods that are "nested":
_resolveCollisions
append
insert
Now append method returns the resid that has been removed due to collisions. I made it because I need to store the removed atoms resid for some applications.

### DIFF
--- a/htmd/molecule/molecule.py
+++ b/htmd/molecule/molecule.py
@@ -298,7 +298,11 @@ class Molecule:
             raise NameError('Failed to insert/append molecule at position {} with error: "{}"'.format(index, err))
 
         if collisions:
-            _resolveCollisions(self, occ1, occ2, coldist)
+            resid_overlaping = _resolveCollisions(self, occ1, occ2, coldist) # store resid collisioning
+            return(resid_overlaping)
+        else:
+            return(None)
+
 
     def remove(self, selection, _logger=True):
         """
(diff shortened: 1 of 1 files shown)


j3mdamas:
hi @cotufalaguera,
thanks for the contribution.

there's a reason `_resolveCollisions` is a protected method. I think that the methods available to the user like `append` should not, by principle, be returning outputs of protected methods, but maybe I'm wrong here. Also, puts the focus of append on the collisions. If somehow we made _resolveCollisions a different type of method, maybe it could be different.

I've put @stefdoerr to review it, so he can give his opinion on that.

stefdoerr:
Ok will check it next week.

On June 1, 2017 5:46:38 AM GMT-05:00, "João M. Damas" <notifications@github.com> wrote:
>hi @cotufalaguera,
>thanks for the contribution.
>
>there's a reason `_resolveCollisions` is a protected method. I think
>that the methods available to the user like `append` should not, by
>principle, be returning outputs of protected methods, but maybe I'm
>wrong here. Also, puts the focus of append on the collisions. If
>somehow we made _resolveCollisions a different type of method, maybe it
>could be different.
>
>I've put @stefdoerr to review it, so he can give his opinion on that.
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/Acellera/htmd/pull/355#issuecomment-305458065

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

j3mdamas:
@stefdoerr is going to write a `detectCollisions` method that returns that, because we don't think that the append method should return that.

stefdoerr:
Added two new functions in `molecule.py`
```python
            idx1, idx2 = _detectCollisions(self, self.frame, mol, mol.frame, coldist)
            torem, numres = _getResidueIndexesByAtom(mol, idx2)
```
Also refactored the old collision detection code which was pretty bad to one 3 times faster and without using atomselect. https://github.com/Acellera/htmd/commit/0ad5baac3e88820428df93d210928bb79b5dba35

This PR is not needed.

stefdoerr:
Thank you in any case for the idea and contribution :smiley:

PR #368, opened 2017-06-22
Tests and Travis fixes for new paradigm

j3mdamas:
* Removes some test files which are outdated/deprecated
* Should fix the problems on PR Travis builds, but avoiding using the secure environment variables or having a fail-safe for them:
    * Only clones htmdlib of not PR, and if credentials do not work for cloning, it carries on with the DSOs 
      on the repo
    * HTMD_NONINTERACTIVE explicitly set, LICENSE_TOKEN removed


j3mdamas:
I removed more tests and added the MacOS lib files (they were missing in the repository, so the PR failed for Mac). You need to re-approve :)

stefdoerr:
You need to fix the run_other_tests.sh script

j3mdamas:
Yeah, yeah, will do tomorrow...

j3mdamas:
Done, can be merged when you approve it.

PR #373, opened 2017-06-26
added spherical coordinate metric for Alberto

stefdoerr:


j3mdamas:
It seems good to me. @cuzzo87 is going to test it with his trajectories and then we merge it. Thanks!

cuzzo87:
I will test it this afternoon. Thx

cuzzo87:
First set of testing, suggests that the arctan function should be arctan2. I will tried and in case fix it

j3mdamas:
I thought we're supposed to only merge if this was working? @stefdoerr?

stefdoerr:
Oops. That was an accident. It must have remembered me accepting it and auto-merged it now that I changed the rules. Check my mail

j3mdamas:
I saw it. It's OK. Let's see if this can be improved in the future

stefdoerr:
@cuzzo87 so was it arctan2? Cause we never fixed it on HTMD

cuzzo87:
I had a couple of test with both. Based on them I would use arctan2.

PR #375, opened 2017-06-26
PullRequest amber and charm build

cuzzo87:
The PullRequest contains modified amber and charm builder files in which were added default ion parameters. The htmd-data files were updated in order to pass the test with the above modifications.

Furthermore non canonical (post translational mutate) residues were added for amber Force Field


stefdoerr:
You did this manually? So for new amber/charmm versions we will have to redo it?

stefdoerr:
Ah I see it's mostly file additions for weird residues. I'm fine with that.

PR #379, opened 2017-06-28
moved init to ui

stefdoerr:
Tests run fine.
Tutorials might not. Have not executed them yet


stefdoerr:
One more reason to hate doctests: If you want them to run without imports you need to add all imports before calling the doctest method. But pycharmm doesn't tell you what imports are missing so you need to either go through them one by one or import *

stefdoerr:
Wanna review Joao or I just merge?

PR #380, opened 2017-06-29
Fix an import in AdaptiveMD class

jeiros:
Hi, I believe I found a small bug. Setting the `save` attribute of the `AdaptiveMD` class to `True` gives an error inside the `_createMSM` method because `os` is not imported alone.


stefdoerr:
Thanks!

stefdoerr:
This is probably due to the latest changes to the `__init__.py`.

PR #382, opened 2017-07-05
Cleaned up file reading in Molecule

stefdoerr:
All read functions now return a Topology and a Trajectory object.
All functions also take same arguments.

This allows us to have multiple readers for the same extension (i.e. .top which is both GROMACS and AMBER topologies, or .crd which can be CHARMM or ...something else) irrespective of the kind of file it is: topology / trajectory etc.

Also much more readable code now.


j3mdamas:
Travis is failing like crazy on these, maybe unrelated?

stefdoerr:
It's been failing since yesterday but due to conda problems as far as I can
tell

But this PR probably also has bugs considering that it's a change to the
deep guts of HTMD

j3mdamas:
As such, this really needs to pass Travis.

stefdoerr:
Which is indeed why I haven't merged yet :P And why I put you two as reviewers, haha

stefdoerr:
Ok, it passed the tests (minus OSX failing as always).
I would probably just wait until Friday to merge so that you can make a safe release tomorrow?

j3mdamas:
Let's see how the last build goes. I will merge it, it's OK.

PR #383, opened 2017-07-06
Adaptive sampling tutorial + new documentation

j3mdamas:
This PR adds:

1. a fixed adaptive sampling tutorial, which is runnable (still to be added to the test-tutorials)
1. moves docs to the new logic (old ones kept in olddoc for now)
1. Improves a bit the documentation for developers

I think it's all

PR #384, opened 2017-07-06
Redoing PR #383

j3mdamas:
Reference: #383


htmd/__init__.py
```diff
@@ -1,11 +1,17 @@
 from htmd.ui import *  # TODO: This will be deprecated in a few versions
-from htmd.config import config
```
stefdoerr: why are you removing config from __init__? It's actually necessary there. cannot be moved to ui

htmd/__init__.py
```diff
@@ -1,11 +1,17 @@
...
 from htmd.version import version as _version
+from htmdx.cli import check_registration, show_news
+from htmd.latest import compareVersions
 import os
 
 config()
 __version__ = _version()
 
+if not (os.getenv("HTMD_NONINTERACTIVE")):
```
stefdoerr: Also registration check should be in ui to avoid unnecessary slowdown and messages when importing directly

stefdoerr:
If you undo the changes to __init__ and ui I'll accept it :)

htmd/__init__.py
j3mdamas: This was a mistake, already fixed it

PR #404, opened 2017-07-21
Factoring out setPhi and getPhi

raimis:
- [x] Fix #393
- [x] Replace `setPhi` with `Molecule.setDihedral`
- [x] Replace `getPhi` with `Molecule.getDihedral`

PR #423, opened 2017-08-07
Parameterize regression tests

raimis:
- [x] Documentation test
- [x] Output directory test
- [x] Parametrization
  - [x] GAFF
  - [x] GAFF2
  - [x] CGenFF
  - [x] RTF and PRM
- [x] Minimize test
  - [x] Run QM
  - [x] Restart
- [x] ESP charge test
  - [x] Run QM
  - [x] Restart
- [x] Dihedral fitting test without optimization
  - [x] Run QM
  - [x] Restart
- [x] Dihedral fitting test with optimization
  - [x] Run QM
  - [x] Restart
- [x] Full parametrization test of benzamidine -- depends on #429 
  - [x] Run QM
  - [x] Restart
  - [x] Frozen charge
  - [x] Select dihedral


raimis:
This depends on #429

raimis:
@j3mdamas, this is read to merge!

This is a set regression tests for `parameterize` covering almost all possible usage (except Queues). It takes ~20 min, but on Travis I trimmed down to <5 min.

j3mdamas:
OK. What about #404? Is it necessary for this merge?

raimis:
No, #404 will go after this!

raimis:
Ping @j3mdamas

j3mdamas:
OK, this is my idea on it:

* Let's not test the log of the parameterize. I think the log is variable and can change as we change the client, and it is really not relevant. The important things to test are the QM data and the parameters generated.
* Let's not put the input string on the data. Command names change, flags change, and we don't want to be releasing htmd-data on those occasions. Let's put that on the code of the tests (which I think makes more sense).
* Let's not double the data by having data for running locally and on Travis. I believe that, by excluding the log, the problem of numerical differences will be automatically solved.

j3mdamas:
I'm doing the mentioned changes

raimis:
@j3mdamas 

- The parameter files are more noisy than the logs. The currently used optimize seems to be unstable.
- I have already ditched or changed some test.
- Don't waste your time the code and the test won't stay for long in that form!

j3mdamas:
I just want to reduce the amount of data in htmd-data and keep the repo as light as possible.

I'll make a proposal on this PR, and we evaluate.

j3mdamas:
I have my proposal, let's see how it goes with Travis

htmd/parameterization/test.py
```diff
@@ -7,235 +7,200 @@
...
-        self.assertTrue('stdout' in paths)
-        #self.assertTrue('stderr' in paths)
-
-        for path in paths:
-            refFile, resFile = os.path.join(refDir, path), os.path.join(resDir, path)
-
-            with self.subTest(path=path):
-                self.assertTrue(os.path.exists(refFile))
+        arguments = cmd.split()
+        returncode = call(arguments, cwd=resDir)
```
raimis: Even if we don't test logs, it would be useful to see stdout and stderr to understand crashes, etc.

htmd/parameterization/cli.py
```diff
@@ -67,18 +67,24 @@ def main_parameterize(arguments=None):
     import numpy as np
     import math
 
-    def printEnergies(m):
-        print("\n == Diagnostic Energies == ")
+    def printEnergies(m, filename):
+        fener = open(filename, "w")
```
raimis: The file isn't closed!

j3mdamas: Fixed in https://github.com/Acellera/htmd/pull/423/commits/76e524779a1b9e79ca770ab0289b59a4991e2488

PR #392, opened 2017-07-18
Parametrize: fix Amber improper reading from a prepi file

raimis:
Fixes partially #391.

PR #429, opened 2017-08-14
Parameterize: add AMBER impropers for the dublicate atom types

raimis:
This is an addition to #392, but it was missed due to lack of testing!


raimis:
@j3mdamas could you review this then you are back, please!

Issue #428, opened 2017-08-13
Add a means for removing hydrogens before parameterizing (because of Amber/tleap)


dr-nate:
One of the many unexpected gotchas when using tleap is that it insists on adding the hydrogens itself, and especially if there are hydrogens on the input structure that do not conform to the Amber naming conventions it will fail. While many files from the PDB do not have hydrogens, some do (NMR structures, or those added by zealous crystallographers). And structures might come from other structural biology tools before coming into HTMD.

So, I propose adding a means to strip all the hydrogens from a structure somewhere. This could be a `True/False` option for the [amber.build(...)](http://software.acellera.com/docs/latest/htmd/htmd.builder.amber.html) method, or perhaps part of the [protein prep](http://software.acellera.com/docs/latest/htmd/htmd.builder.preparation.html) module, or part of [molecule module](http://software.acellera.com/docs/latest/htmd/htmd.molecule.molecule.html). I didn't see it implemented anywhere but correct me if I'm wrong.

On the one hand, it might seem best to just add it as something like `strip_hyrogens=True` to the `amber.build()` method, but then the user might end up with a protonation state they didn't expect unless there's a nice big warning. If you add it as a method elsewhere, it will be more versatile, but then the user will have to know to run it before feeding the structure into the amber build. Or it will have to be called inside the build method, which might be the best of both ideas.

Just my thoughts on the issue. If I have time I'll implement it, but currently just going to use `awk '{if ($3 !~ /^H/ ) print $0}' in.pdb > out.pdb` and go to the next step.

stefdoerr:
You can of course do `mol.filter('noh')` which will remove all hydrogens from a Molecule.

I would not remove during build because for example CHARMM doesn't modify resnames for protonated residues so once you remove the hydrogens there is no way of going back.

j3mdamas:
Hi Nate,

the `proteinprepare` does not solve your problems? Could you attach here an example of what you are facing so we may look at it?

Thanks!
João

dr-nate:
Stefan's suggestion of using the filter method works fine. The lingering question is whether you want to hand-hold the user by adding the functionality I suggested. I'm actually a little ambivalent as well because on the one hand this would be somewhat hidden functionality that might mess up the user's intent, but on the flip side tleap errors are really cryptic and so this could really save them some time.

[...]

PR #430, opened 2017-08-16
Add automatic hydrogen removal for amber build function.

dr-nate:
Added a `strip_hydrogens=True` argument to the build function in `htmd.builder.amber` so users can avoid chasing down a common issue in tleap. A possible solution to #428.


stefdoerr:
@j3mdamas what do you think of this? I'm fine with the option but making it True on default would change behaviour in HTMD. On the other hand putting it False will mean no-one sees/uses it.
I'm kind of leaning towards changing the default behaviour.

j3mdamas:
I disapprove of this PR. I think that rather going in one easy way (which also introduces a flag/variable that starts to fork the attempt at equivalency between `amber.build` and `charmm.build`), we should try to address the actual problem.

As I said before on the issue, we circumvent this issue by passing the structure through `proteinPrepare`. Is it a problem that existing hydrogens names are not changed to the correct ones by that function?

We have clients that do not want the hydrogens to be changed from what they provide, so that adds up to me not approving this PR.

j3mdamas:
I'll close it, and let's get back to the issue #428

PR #443, opened 2017-09-05
Refactoring torsion parameter fitting

raimis:
- [x] Code
  - [x] Clean `FFMolecule`
  - [x] Move out QM from the fitting loop
  - [x] Move fitting code into a separate class
  - [x] Generalize objective function
  - [x] Use *NLopt* for optimization
- [x] Tests
  - [x] Revive disabled `parameterize` fitting tests
  - [x] Add more `parameterize` fitting tests
  - [x] Fully functional `FakeQM`
  - [x] Unit tests
- [x] Documentation

Depends on #437


j3mdamas:
I hope there can be a rebase of this branch after we merge #437 :smile:

htmd/parameterization/cli.py
```diff
@@ -51,20 +51,28 @@ def cli_parser():
                         default=None, dest="freezeq")
     parser.add_argument("--no-geomopt", help="Do not perform QM geometry optimisation when fitting torsions (default: "
                                              "%(default)s)", action="store_false", dest="geomopt", default=True)
+    parser.add_argument("--seed", help="Random number generator seed (default: %(default)s)", type=int,
```
j3mdamas: Is this useful for the user? I would add `help=argparse.SUPPRESS`

htmd/parameterization/cli.py
```diff
@@ -366,7 +334,6 @@ def printEnergies(m, filename):
                 f.close()
             except ValueError as e:
                 print("Not writing Amber FRCMOD: {}".format(str(e)))
-    sys.exit(0)
```
j3mdamas: Why removing the `sys.exit`?

htmd/parameterization/ffmolecule.py
```diff
@@ -261,518 +243,117 @@ def getDipole(self):
...
-        self.qm.optimize = geomopt
-        self.qm.restrained_dihedrals = np.array(frozens)
-        self.qm.directory = fitdir
-        results = self.qm.run()
+            self.qm.molecule = molecule
+            self.qm.esp_points = None
+            self.qm.optimize = geomopt
+            self.qm.restrained_dihedrals = np.array([dihedral])
+            self.qm.directory = fitdir
+            qm_results.append(self.qm.run())  # TODO submit all jobs at once
```
j3mdamas: Doesn't this mean that each rotamer is done in a serial way? This needs to be changed before merging.

htmd/parameterization/cli.py
raimis: Why do you need it at the end of the function.

htmd/parameterization/cli.py
raimis: Yes, it is!

htmd/parameterization/cli.py
j3mdamas: The `main_parameterize()` function is the one that is executed by the command line. As such, we have a `sys.exit(0)` at the end of it

raimis: Exit status is set to 0 implicitly, if the there is no exceptions raised.

htmd/parameterization/ffmolecule.py
raimis: No! QM class already parallelize over rotamers. The remaining issue, it is not parallized over dihedrals, i.e. QM jobs will be run in batches of 36, but not all at once.

Until AceCloud is available, we can live with that.

j3mdamas: ok ok

htmd/parameterization/cli.py
j3mdamas: ok, it was there just to be safe

htmd/parameterization/cli.py
j3mdamas: As in, if a parameterization fails, they could try a different seed?

raimis: Exactly! :)

htmd/parameterization/dihedral.py
```diff
@@ -0,0 +1,548 @@
...
+        types = [self.molecule._rtf.type_by_index[index] for index in dihedral]
+
+        all_uses = []
+        for dihedral_indices in self.molecule.dihedrals:
+            dihedral_types = [self.molecule._rtf.type_by_index[index] for index in dihedral_indices]
+            if types == dihedral_types or types == dihedral_types[::-1]:
+                all_uses.append(dihedral_indices)
+
+        # Now for each of the uses, remove any which are equivalent
+        unique_uses = [dihedral]
```
j3mdamas: `unique_uses` seems like a confusing variable name for a list of equivalent dihedrals

htmd/parameterization/dihedral.py
```diff
@@ -0,0 +1,548 @@
...
+        if len(self.dihedrals) != len(self.qm_results):
+            raise ValueError('The number of dihedral and QM result sets has to be the same!')
+
+        # Get dihedral names
+        self._names = ['%s-%s-%s-%s' % tuple(self.molecule.name[dihedral]) for dihedral in self.dihedrals]
+
+        # Get equivalent dihedral atom indices
+        self._equivalent_indices = []
+        for idihed, dihedral in enumerate(self.dihedrals):
+            for rotableDihedral in self.molecule._soft_dihedrals:
```
j3mdamas: Let's try to make nomenclature simple: from what I know, soft and rotatable torsions are equivalent terms in the parameterize context. Just use one.

htmd/parameterization/dihedral.py
```diff
@@ -0,0 +1,548 @@
...
+        opt = nlopt.opt(nlopt.LN_BOBYQA, opt.get_dimension())
+        opt.set_min_objective(self._objective)
+        opt.set_lower_bounds(lower_bounds)
+        opt.set_upper_bounds(upper_bounds)
+        opt.set_xtol_rel(1e-3)
+        opt.set_maxeval(neval)
+        opt.set_initial_step(1e-3*(upper_bounds-lower_bounds))
+
+        # Optimize parameters
+        vector = opt.optimize(vector) # TODO check optimizer status
```
j3mdamas: Just some terms inconsistency between this optimization and the random_search one, perhaps on purpose, just checking: in one you use the vector, loss generic terms, while in the other you use best_vector, best_loss.

htmd/parameterization/ff.py
```diff
@@ -202,16 +202,11 @@ def writeFrcmod(self, rtf, filename):
...
-                    sign = 1
+                for p in prm:
                     scee = 1. / p.e14
                     scnb = 2.
-                    if pi < (len(prm) - 1):
-                        sign = -1
                     print("%2s-%2s-%2s-%2s 1 %12.6f %12.6f %12.6f %12.6f %12.6f" %
-                          (t[i.types[0]], t[i.types[1]], t[i.types[2]], t[i.types[3]], p.k0, p.phi0, sign * p.n, scee,
-                           scnb),
+                          (t[i.types[0]], t[i.types[1]], t[i.types[2]], t[i.types[3]], p.k0, p.phi0, p.n, scee, scnb),
```
j3mdamas: What happened to the sign? No longer needed?

raimis: I have no idea, why it was needed at the first place.

htmd/parameterization/dihedral.py
raimis: Changed!

htmd/parameterization/dihedral.py
raimis: I'm uniforming names, but this comes from not yet re-factored part.

PR #444, opened 2017-09-05
Out-of-memory HDF5 MetricData

stefdoerr:
Allows us to work with huge datasets (albeit a bit slower, depending on HDD/SSD speed)

TODO:

- [ ] Make TICA work with it (needs streaming TICA if hdf5 type is detected)
- [ ] Find out if it's possible to store objects (Sim) alongside HDF5

### DIFF
--- a/htmd/metricdata.py
+++ b/htmd/metricdata.py
@@ -124,7 +124,7 @@ class MetricData:
         Centers of clusters
     """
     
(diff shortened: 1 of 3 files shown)


stefdoerr:
And yes @giadefa this is related to protein folding :yum: (synthetic adaptive specifically)

j3mdamas:
I saw some HDF5-stuff lately on PyEMMA repo. Maybe something interesting for this?

stefdoerr:
From a quick search it seems the only use HDF5 for storing trajectories for random access

PR #453, opened 2017-09-25
Queue properties

j3mdamas:
Uniformization of SimQueues.  …
   * Creation of (abstract) generic properties that can be probed for any implemented SimQueue: ncpus, ngpus, memory
   * Generalization of the wait method (abstract method on SimQueue now), to include sentinel files and notcompleted mode (all in SimQueue)
   * super __init__ call replaced by separate calls to each parent, as it was not working (when I created an __init__ for SimQueue)
   * LocalQueues:
        - run_job converted into method
        - _getmemory method implemented
        - ncpus of LocalQueue detached from the ncpus of htmd

 TODO:
     - Merge LocalGPUQueue and LocalCPUQueue into a single class which manages both ncpu and ngpu simultaneously
     - Implement generic properties and notcompleted on AceCloudQueue

### DIFF
--- a/htmd/queues/localqueue.py
+++ b/htmd/queues/localqueue.py
@@ -51,6 +51,7 @@ class _LocalQueue(SimQueue, ProtocolInterface):
             self._queue = queue.Queue()
 
             devices = self._getdevices()
+            self.memory = self._getmemory()
 
             self._threads = []
             for d in devices:
@@ -227,9 +228,11 @@ class LocalGPUQueue(_LocalQueue):
     ngpu : int, default=None
         Number of GPU devices that the queue will use. Each simulation will be run on a different GPU. The queue will
         use the first `ngpu` devices of the machine.
-
     devices : list, default=None
         A list of GPU device indexes on which the queue is allowed to run simulations. Mutually exclusive with `ngpu`
+    memory : int, default=None
+        The amount of RAM memory available for each job. If None, it will be guessed from total amount of memory and
+        the number of devices
 
 
     .. rubric:: Methods
@@ -249,7 +252,8 @@ class LocalGPUQueue(_LocalQueue):
         self._arg('devices', 'list', 'A list of GPU device indexes on which the queue is allowed to run '
                                      'simulations. Mutually exclusive with `ngpus`', None, val.Number(int, '0POS'),
                   nargs='*')
-        self._arg('memory', 'int', 'The amount of RAM memory available for each job.', self._getmemory(),
+        self._arg('memory', 'int', 'The amount of RAM memory available for each job. If None, it will be guessed from '
+                                   'total amount of memory and the number of devices', None,
                   val.Number(int, '0POS'))
 
     def _getdevices(self):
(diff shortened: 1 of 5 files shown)


htmd/queues/localqueue.py
```diff
@@ -117,7 +154,9 @@ def submit(self, mydirs):
         """
         self._setupQueue()
 
-        if isinstance(mydirs, str): mydirs = [mydirs]
+        if isinstance(mydirs, str):
+            mydirs = [mydirs]
+        self._dirs.extend(mydirs)
```
stefdoerr: why here and not in .submit?

htmd/queues/localqueue.py
```diff
@@ -144,12 +183,37 @@ def inprogress(self):
...
     def _getdevices(self):
-        pass
+        return list()
+
+    def _getmemory(self):
+        ureg = UnitRegistry()
+        total_memory = int(ureg.Quantity(psutil.virtual_memory().total, ureg.byte).to('MiB').magnitude)
+        nr_devices = len(self._getdevices())
+        if nr_devices != 0:
+            return int(total_memory/nr_devices)
```
stefdoerr: This is horribly nasty btw. It assumes the computer has 0 memory in use.

j3mdamas: virtual_memory().total is all total memory available, not only the available.

what do you propose?

htmd/queues/localqueue.py
j3mdamas: oh right, this is a bug, I'll change it.

stefdoerr: ah forget it. bad diff viewer

stefdoerr:
a) that does not tell me why they have to be properties. You could make them normal "args" of the class like the rest of the options, no?
b) yes I am actually proposing on relying on disk information but keeping tabs of simulations using a disk-based database

j3mdamas:
a) as properties, you can oblige them to be defined in child classes, and also standardizes naming (queue vs. partition problem).
b) sure!

stefdoerr:
Ah yes ok for the standardization. right. Otherwise it's two separate
variables
Ok I'm fine with it

raimis:
@j3mdamas @stefdoerr 

`LocalCPUQueue` is still broken:
```python
>>> from htmd.queues.localqueue import LocalCPUQueue
>>> q = LocalCPUQueue()
>>> print(q.memory)
None
>>> print(q.ncpu)
None
```
Some CPUs can be found! 
```python
>>> q._getdevices()
[None, None, None, None, None, None, None, None]
```
Well, I expected a bit more...
```python
>>> q._getmemory()
>>> 1978
```
```bash
$ free -m
              total        used        free      shared  buff/cache   available
Mem:          15830        2961        9031         702        3837       11773
Swap:         16383           0       16383
```

stefdoerr:
Works as expected I'd say.
ncpu is not available cpus. It's how many you have set (which you haven't)

stefdoerr:
Same for `q.memory`

stefdoerr:
Just look at the code really.
```python
    def _getmemory(self):
        ureg = UnitRegistry()
        total_memory = int(ureg.Quantity(psutil.virtual_memory().total, ureg.byte).to('MiB').magnitude)
        nr_devices = len(self._getdevices())
        if nr_devices != 0:
            return int(total_memory/nr_devices)
        else:
            return None
```
As you can see it divides the memory by number of devices.
I don't know if I agree with this way, I'd personally report whole memory and then divide outside the function but that's up to @j3mdamas

j3mdamas:
Just fixed it with @raimis. I've changed the `LocalCPUQueue` to have defaults.

PR #457, opened 2017-09-27
sequenceStructureAlignment modified

cuzzo87:
The function was modified in order to be able to use more than one piece of sequence to perform an alignment. 
One argument added which default value is 1, so the default behavior is exactly the same as before.


stefdoerr:
You commited ATP frcmod in the same PR @cuzzo87  :(

stefdoerr:
Otherwise it's great. But remove those files from the PR (or remake the PR, don't know what's easier)

stefdoerr:
I still see 7 files changed in the PR :( I think it might really be easier to make a new PR with just the file

cuzzo87:
Ok I wil do it from scratch. I am sorry for this.

PR #459, opened 2017-09-27
Added orient function for principal axes

tonigi:


stefdoerr:
Is this like the PCA orientation I did in OPM paper to fit the proteins better into the box? I don't know if I'd make it a method of Molecule instead of a util function

stefdoerr:
In the sense that it doesn't feel like a very important feature for Molecules and we don't want to spam them full with methods. So stuff like that goes usually to `util.py`

tonigi:
Ok moving it away. Kill the PR. The problem with util, if I understand, is that they do not modify the object in place (an asymmetry which I find confusing).

stefdoerr:
Mmm yes. Personally I find it intuitive. Class methods modify object in-place. Functions in python are usually designed to return results since there is no limit to number of returning variables like in C. I find in-place modification in python functions a bit sneaky (although I'm sure I might have done it somewhere, but I regret it)

tonigi:
Yes, I also prefer returns, which allow things like x.cross(y).normalize() . My issue is that now we have a mixture of the two styles.

stefdoerr:
Yes I see the point but class methods returning objects at each operation though makes for some real mess and waste of memory. For methods which you want to call 1000 times I prefer to keep the basic operations as in-place modifications.

PR #462, opened 2017-09-28
Add implementation for sequenceStructureAlignment

cuzzo87:
the same of #457 but correct the wrong commitments


stefdoerr:
Did they upgrade travis? We are passing tests in 15 minutes!

PR #468, opened 2017-10-03
Fix local queue

cuzzo87:
Import os error while using Local queue. I have just restored the import on the top,


cuzzo87:
Stefan. Sorry I have seen now the two not related commits. Do not merge. I will do it again.

PR #477, opened 2017-10-16
improved "read parameter card" compatibility

ismaelresp:
The real synstax in CHARMM is:
READ PARAmeter CARD
Upper case: mandatory
Lower case: optional

### DIFF
--- a/htmd/builder/charmm.py
+++ b/htmd/builder/charmm.py
@@ -754,7 +754,7 @@ def split(filename, outdir):
     outprm = os.path.join(outdir, 'par_{}.prm'.format(base))
 
     startrtf = re.compile('^read rtf card', flags=re.IGNORECASE)
-    startprm = re.compile('^read param? card', flags=re.IGNORECASE)
+    startprm = re.compile('^read para(m|me|met|mete|meter)? card', flags=re.IGNORECASE)
     endsection = re.compile('^end', flags=re.IGNORECASE)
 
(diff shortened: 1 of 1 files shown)


j3mdamas:
@ismaelresp let me know when you're up in the lab

ismaelresp:
@j3mdamas i am at the lab now

j3mdamas:
I have a meeting now, I'll be there 1 hour or so. Is that OK?

ismaelresp:
@j3mdamas Yes, it's OK. I'll be in my workplace.

stefdoerr:
I replaced it with `'^read para* card'` for cleanliness' sake. Thanks for the PR though!

j3mdamas:
@stefdoerr your replacement works but it's not correct synthax. Example, if there's a an entry called `read paraXYZXYZ card`, it will be accepted by your synthax, but is not the official synthax. @ismaelresp's contribution, albeit not very elegant, was the real interpretation of the CHARMM synthax (which, if you ask me, it's pretty absurd).

stefdoerr:
Yeah but really I don't think we care if we over-match with the regex as long as we don't undermatch. There is pretty much no real-case scenario where it would say `read paraXYZHELLOJOAO card` :D

j3mdamas:
I agree with you, I was just stating that technically it was not according to the absurd CHARMM synthax :)

PR #478, opened 2017-10-16
Stream file names no longer need to start with "toppar_[alfanum]+".

ismaelresp:
This is usefull to use parameters just as they come out of paramchem together with other cases. I think that it is not necessary to make the user rename all their str files  (I have 60 systems with different ligands) and there is no need for the "toppar_[alfanum]+" requirement.


stefdoerr:
Good idea. Sorry for not seeing this earlier.

PR #483, opened 2017-10-19
Fixing _guessMass and refactoring vdw module

j3mdamas:
@stefdoerr could you make sure I did not screw voxeldescriptors?
@raimis could you make sure that I did not screw parameterize?

thank you both


htmd/molecule/voxeldescriptors.py
```diff
@@ -259,9 +259,9 @@ def _getRadii(mol):
             raise ValueError('PDBQT element {} does not exist in mappings.'.format(elem))
         elem = mappings[elem]
 
-        if elem in VDW.elements:
-            rad = radiusByElement(elem)
-        else:
+        try:
```
stefdoerr: no, thats just ugly. why try catch when you can check?

stefdoerr: Or check in the function instead of here

j3mdamas: sorry, stupidity from my side :)

j3mdamas:
@stefdoerr do you approve?

PR #486, opened 2017-10-24
Parameterize: set the numbers of CPU and GPU

raimis:
Even if we set the *Slurm* partition to *normalCPU*, it still wants GPUs.


raimis:
@j3mdamas, this is a bit hackish. Any better idea?

j3mdamas:
This is already fixed by a commit I'm making. I'll close this PR.

PR #497, opened 2017-11-07
Attempt at fixing restart of QM jobs on any SimQueue

j3mdamas:
@raimis what do you think? It may not be ideal, but at least it's better than having the restart only working on `LocalCPUQueue`


htmd/qm/base.py
```diff
@@ -160,13 +160,11 @@ def _submit(self):
...
                 self.queue.submit(directory)
+        self.queue._dirs = self._directories
 
     def _retrieve(self):
 
-        if isinstance(self.queue, LocalCPUQueue):
-            self.queue.wait(sentinel=False) # TODO: sentinel is still broken
-        else:
-            self.queue.wait(sentinel=True)
+        self.queue.wait(sentinel=True)
```
raimis: According to the failed tests, `LocalQueue` with `sentinel=True` is still broken. Two weeks  ago I was looking into that, but no clue what is wrong... :(

htmd/qm/base.py
```diff
@@ -160,13 +160,11 @@ def _submit(self):
         for directory in self._directories:
             if not self._completed(directory):
                 self.queue.submit(directory)
+        self.queue._dirs = self._directories
```
raimis: It is not the best idea to modify a private attribute out side of the object.

htmd/qm/base.py
j3mdamas: I made it work on my machine, I can't understand what is going on on Travis. Only two tests fail.

htmd/qm/base.py
j3mdamas: I'm willing to make that compromise in order for people who run with Lsf or Slurm being able to restart a parameterize job, which is not possible currently

htmd/qm/base.py
raimis: `LocalQueue` is using threads, so we might have the rate race as it behaves differently. The failing tests are the ones, which actually run QM, not just restart.

j3mdamas: The failing tests are:
test_h2o2_min and test_h2o2_esp in parameterization/test.py
test_restrained_dihedrals in qm/test.py
I'm really puzzled on why they are failing on Travis but passing on my machine

j3mdamas: rate race?

j3mdamas: the first two ones are single QM jobs, how is there a race?

htmd/qm/base.py
raimis: I don't get how this helps. Isn't this make to restart all simulation, even the ones, which already finished successfully?

It just copy `_directories` to `_dirs`, so it not checked with `_completed` method. Or am I missing something?

htmd/qm/base.py
raimis: There is an additional thread create even for a single job, so it can interfere with the main thread.

htmd/qm/base.py
j3mdamas: It will not matter, as the completed ones should also have sentinel files.
By copying the `_directories` into `_dirs`, the `notcompleted` method actually works instead of failing due to having the `_dirs` being empty (which is a RuntimeError in general runs, but not on a restart run). We never devised the queues to manage the restart of simulations (MD or QM). We may want to make it more general and implemented in that way, but right now we need a fix

htmd/qm/base.py
j3mdamas: Let's fix this tomorrow.

htmd/qm/base.py
raimis: I see. If you just remove `if not self._completed(directory)`, all directories would be added unconditionally by `submit`. And this will avoid modifying the private variable directly.

j3mdamas: Ok, yes, it's true. The problem is that submitting rewrites inputs, cleans sentinels and will actually launch the jobs again, not caring about the output that is already there.

j3mdamas:
Any advances?

j3mdamas:
Closing this PR

PR #502, opened 2017-11-09
Parameterize: fix atom type handling with "x"

raimis:
Duplicated atom types are named original_name + "x" + number, e.g. `ca` --> `cax0`. Fix the case if an atom type contains "x", so `cx` --> `cxx0`.


j3mdamas:
could you please do this fix on `rel-1.10.x` branch and do PR to there? thanks!

PR #504, opened 2017-11-09
Queue failures improvements

### LINKED ISSUE #491: Queue needs to manage scratch!
By default, *Psi4* uses `/tmp` for anything that does not fit into RAM. In case of our own infrastructure, it has be changed to a different path. Ideally, a queue object should handle that and pass to *Psi4*.

We need something like that working:
```
>>> queue.scratch
'/huge_disk/tmp'
```

j3mdamas:
- [x] Failures of submission do not inform the user nicely about the reason for failure (normally the submission commands throw to stdout/stderr the reason the job was not submitted)
- [ ] Management of scratch/temporary space (create variables in Queues, manage those variables appropriately, make relevant applications (Psi4, Gaussian) use those variables, cleaning of the temporary files of any exit, successful or unsuccessful) (Fixes #491)
- [ ] Creation of sentinel file on any exit, successful or unsuccessful (solves hangs of processes requiring sentinel files; if cleaning is figured out on the point above, just replicate)

### DIFF
--- a/htmd/queues/lsfqueue.py
+++ b/htmd/queues/lsfqueue.py
@@ -8,7 +8,7 @@ import shutil
 import random
 import string
 import numpy as np
-from subprocess import check_output, CalledProcessError
+from subprocess import check_output, STDOUT, CalledProcessError
 from protocolinterface import ProtocolInterface, val
 from htmd.queues.simqueue import SimQueue
 from htmd.config import _config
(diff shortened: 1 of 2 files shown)


stefdoerr:
Nice! Makes sense, I like the improvements :)

j3mdamas:
I'm closing this PR and rather opening an issue on it (#610)

PR #519, opened 2017-12-11
Parameterize: full dihedral angle parallelization

raimis:
- [x] Refactor QM classes
- [x] Simultaneous QM job submission

Depends on #521 
Fixes #517

PR #527, opened 2017-12-18
Parameterize: improve restartability

raimis:
- [x] Fix restartability with non-local queues
- [x] More robust detection of completed vs killed Psi4 jobs
  - [x] Absuse `timer.dat`
  - [x] Update tests

Depend on #519


j3mdamas:
Is it duplicate of this https://github.com/Acellera/htmd/pull/497? Do we deprecate and close the other PR?

j3mdamas:
Why aren't we sticking with the `htmd.queues.done`, which is general, and relying on the `timer.dat`?

raimis:
`htmd.queues.done` will be present regardless, what happened to the job. `timer.dat` is only present, if Psi4 finished successfully or hit unrecoverable error, while random crashes (e.g. memory shortage), user killings can be recovered.

raimis:
@stefdoerr 

One more PR for your review and merging! :D

htmd/qm/psi4.py
```diff
@@ -118,7 +118,10 @@ def _command(self):
         return 'export HTMD_PSI4_WORKDIR=$(pwd)\npsi4 -i psi4.in -o psi4.out &> psi4.log'
 
     def _completed(self, directory):
-        return os.path.exists(os.path.join(directory, 'psi4.out'))
+        # Abuse "timer.dat" to detect if a Psi4 job has completed.
```
stefdoerr: Is it not possible to use sentinels the same way we use them in our queues? By wrapping the executable in a bash script with the sentinel?

stefdoerr: Ah ok I read the discussion

PR #529, opened 2017-12-21
QM: basis set substitution

raimis:
- [ ] Basis set substitution
  - [ ] ??? basis set support
- [ ] Integration with Psi4
- [ ] Tests
- [ ] Documentation


j3mdamas:
Do you think 1.12 is feasible?

raimis:
Yes!

PR #535, opened 2018-01-18
Minor fixes

stefdoerr:


htmd/queues/localqueue.py
```diff
@@ -115,8 +115,10 @@ def _createJobScript(self, fname, workdir, runsh, gpudevice=None):
             # Move completed trajectories
             if self.datadir is not None:
                 datadir = os.path.abspath(self.datadir)
-                if not os.path.isdir(datadir):
+                try:
                     os.mkdir(datadir)
+                except OSError:
```
raimis: Do you try to catch the case that the directory already exists?

stefdoerr: yes

stefdoerr: There is a weird race condition when creating the directory on /tmp/ that Alberto hit upon.

stefdoerr: Well maybe OSError is too broad. I could catch FileExists error

stefdoerr:
Hm now I see how fork PRs work. This PR will track my master until I close it. Maybe I should make branches and PRs then......

htmd/queues/localqueue.py
raimis: Or just use 
```python
os.makedirs(datadir, exist_ok=True)
```
without any exception catching...

htmd/queues/localqueue.py
```diff
@@ -117,7 +117,7 @@ def _createJobScript(self, fname, workdir, runsh, gpudevice=None):
                 datadir = os.path.abspath(self.datadir)
                 try:
                     os.mkdir(datadir)
-                except OSError:
+                except FileExistsError:
                     pass
```
raimis: Or just use 
```python
os.makedirs(datadir, exist_ok=True)
```
without any exception catching...

stefdoerr: Ah yes. I sometimes forget that one exists... Thanks

PR #540, opened 2018-01-31
Parameterize: improve Psi4 and queue handling

raimis:
- [ ] use Psi4 from an active conda environment
- [ ] pass queue memory in correct units


raimis:
@stefdoerr, can you review this, please!

PR #545, opened 2018-02-01
Plotting the FES with different MetricData object. Standardized plotting behaviour

stefdoerr:


cuzzo87:
I agree with all the commits. The last one I cannot give you feedback about the correctness, but I like the new feature implemented

PR #546, opened 2018-02-02
QM: Psi4 uses more conservative memory estimate

raimis:
- [x] Use more conservative memory estimate, otherwise Psi4 exceeds the memory limit of a queuing system


cuzzo87:
Well done. For me it is ok

j3mdamas:
Since they go to the troubles of having a parameter just for this, I guess it's a general problem. Any particular reason for 70%? Seems quite rough! 90% was not enough?

PR #551, opened 2018-02-05
replaced python dihedral with numba dihedral which is 2 orders of mag faster

stefdoerr:
Had to decide if to return rad or deg. I think rad is more standard for programming so I picked that


htmd/molecule/molecule.py
```diff
@@ -724,8 +724,8 @@ def getDihedral(self, atom_quad):
         --------
         >>> mol.getDihedral([0, 5, 8, 12])
         """
-        from htmd.molecule.util import dihedralAngle
-        return np.deg2rad(dihedralAngle(self.coords[atom_quad, :, self.frame]))
+        from htmd.numbautil import dihedralAngle
```
raimis: Why `numbautil` instead just `util`. Does the implementation matters?

stefdoerr: Eh I was collecting all numba implementations of small stuff in a single file.

stefdoerr: If it grows bigger or more general I'll divide it in more files. For the moment I don't care much about the name. There is already a `htmd.util` and I didn't want to add this stuff there.

htmd/numbautil.py
```diff
@@ -23,8 +23,62 @@ def wrapBondedDistance(d, box):
...
+        The size of the periodic box
+    """
+    '''
+    http://en.wikipedia.org/wiki/Dihedral_angle#Methods_of_computation
+    https://www.cs.unc.edu/cms/publications/dissertations/hoffman_doug.pdf
+    http://www.cs.umb.edu/~nurith/cs612/hw4_2012.pdf
+
+    Ua = (A2 - A1) x (A3 - A1)
+    Ub = (B2 - B1) x (B3 - B1) = (A3 - A2) x (A4 - A2)
+    angle = arccos((Ua * Ub) / (norm(Ua) * norm(Ub)))
```
raimis: The equation do not match the actual code!

stefdoerr: Yeah not anymore. It matched the old code. Removed the equation

PR #552, opened 2018-02-05
added file searcher for amber build and all files are copied locally …

stefdoerr:
…in the build dir now.


htmd/builder/amber.py
```diff
@@ -227,8 +267,15 @@ def build(mol, ff=None, topo=None, param=None, prefix='structure', outdir='./bui
...
         ff = [ff]
-    for force in ff:
-        f.write('source ' + force + '\n')
+    for i, force in enumerate(ff):
+        if os.path.isfile(force):
+            f.write('source {}\n'.format(force))
+        else: # Search amber directories for the file
+            p = _locateFile(force, 'ff', tleap)
+            if p is not None:
+                newname = 'ff{}_{}'.format(i, os.path.basename(force))
```
cuzzo87: Is there any reason for prepending ```ff``` in the param filename?

stefdoerr: It's to the forcefield files. Why not? I did it to separate the param from the topo and the ff since I prepend them some numbers.

cuzzo87:
Apart for the comment above is ok. More clear and easy to ready than before

PR #556, opened 2018-02-08
Better parametrize testing

stefdoerr:
I am trying to replace the file comparison in parametrize tests with an actual energy evaluation. 

1. it compares all non-dihedral parameters for exact matching
2. it evaluates the energy of the dihedrals of both parametrizations and calculates the RMSE

TODO:

- [x] Implement it in the tests
- [ ] Remove file comparison for `frcmod` and `prm`
- [ ] Determine the threshold for the RMSE of two fits

### DIFF
--- a/htmd/parameterization/test.py
+++ b/htmd/parameterization/test.py
@@ -140,6 +140,19 @@ def _parameterCompare(folder1, folder2, fields=('atom_types', 'bond_types', 'ang
     return _compareDihedralEnergies(mol, prm1, prm2, dihedrals)
 
 
+def _foldersToTest(refDir):
+    folderstotest = []
+    excluded = ('minimize', 'esp', 'dihedral')
+    for root, dirs, _ in os.walk(refDir, followlinks=True):
+        for d in dirs:
+            dname = os.path.relpath(os.path.join(root, d), start=refDir)
+            if dname in excluded:
+                continue
+            if os.path.exists(os.path.join(root, d, 'mol.frcmod')) or (os.path.exists(os.path.join(root, d, 'mol.prm')) and os.path.exists(os.path.join(root, d, 'mol.rtf'))):
+                folderstotest.append(dname)
+    return folderstotest
+
+
 class TestParameterize(unittest.TestCase):
 
     def setUp(self):
@@ -152,8 +165,7 @@ class TestParameterize(unittest.TestCase):
 
         print('\n') # Just for a better readability
 
-    def _test(self, refDir, resDir, cmd):
-
+    def _execute(self, refDir, resDir, cmd):
         if not os.path.exists(resDir):
             os.makedirs(resDir)
 
@@ -165,6 +177,7 @@ class TestParameterize(unittest.TestCase):
         returncode = call(arguments, cwd=resDir)
         self.assertEqual(returncode, 0)
 
+    def _test(self, refDir, resDir):
         filestotest = []
         excluded = ('minimize', 'esp', 'dihedral', '.coor', '.svg')
         for root, _, files in os.walk(refDir, followlinks=True):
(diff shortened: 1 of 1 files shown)


stefdoerr:
As in this branch all tests are deterministic and no changes were made to the parametrization I cannot determine a good RMSE (it's 0 right now). So if everyone is fine with it I will merge this into `master` and from there merge it to `dihedralnumba` branch where I can see how different my results are.

@raimis if you can give this a look it would be great

j3mdamas:
@stefdoerr this is an incomplete PR, right? Right now, what was merged is just adding some `Molecule`, and "parmed" parameter class comparisons to what was already being done and not actually replacing anything. Can you explain better the merge since I don't understand it from https://github.com/Acellera/htmd/pull/556#issuecomment-364173039?

stefdoerr:
I needed to determine  a threshold for the energy difference when testing. Since this branch would give 0 energy difference as nothing changed in the actual parametrization code I needed to merge it into dihedralnumba class which actually gave divergent results to determine a good threshold. It's not incomplete. It's merged :P

j3mdamas:
You need to explain it to me tomorrow, because I don't understand this PR as it is not achieving what you said in the OP, and I don't understand why what you added (which is great, I'm not criticising the actual code of the PR) gives "0 energy difference".

Is this zero energy difference to pass the tests? If it is, I don't know how it's helping passing them, as the `energies.txt` file is still being compared on the `_test()`.

j3mdamas:
Furthermore, you could have always merged this branch into the "numba" branch, without needing to "pass through" master.

stefdoerr:
Did that last time and ended up duplication commits and making an
unreadable PR.

stefdoerr:
The `energy.txt` gives you total energy difference between the two parametrizations. What I cared about was the RMSE of the two fitted curves to each other (new one versus the one in htmd/data/test-param) to make sure that the fits don't get worse then the old ones. To do that I evaluate old and new parameters at all 36 rotamers and get the rmse. That is what the code does and it is implemented in commit https://github.com/Acellera/htmd/pull/556/commits/b1e1bd31bd3342882c230efa048b3cc13a0ba873#diff-30df3daf0d6dd9196a6be06176a32d20

In this PR if you ran the tests, the parametrization you produce is identical to the one in the tests. Hence the RMSE of the two curves is 0.

j3mdamas:
Sorry Stefan, for a reason I was seeing a wrong diff, where I didn't see `_compareDihedralEnergies`.

j3mdamas:
More correctly, I was overlooking it.

j3mdamas:
Amazing code btw! :D I just think that the PR post should actually reflect what was done in the PR :P 

Just a small critique: the way you did it is not the same as comparing the dihedral plots. The dihedral plots are generated from the energies of the final minimized configurations that can be found on each folder inside each angle of each dihedral, while you are comparing the initial configurations (you are generating them in the same way that is generated for the QM calculation). I guess the way to compare equal to comparing the plots would be to compare the energies given by the final configurations found in `dihedral/<dihedral-name>/<dihedral-angle>`. Or even better would be to output the raw values that are used on the plots in a text file and compare those xD

As a positive part, you test actually tests even when dihedrals are not parameterized.

raimis:
@stefdoerr 

@j3mdamas  is right, it could be done simple outputting plot data and comparing just file.

Another issue, you are using functions for reference data, which are suppose to be tested. So the test are not testing what they are suppose to test.

stefdoerr:
I don't see why write another temp file when I can do the comparisson by evaluation which is more or less the ultimate test. Plus it gives more flexibility.
@j3mdamas Ok I will use the final conformations. Didn't know that
@raimis I don't understand what you mean with functions for reference data.

j3mdamas:
Actually, I don't think the raw data of the plots would be a temporary file, I really think we should output it (and use it, since it's there), as users may want to redo the dihedrals plot some other way. @Acellera/gerard already mentioned that before.

PR #567, opened 2018-02-21
Parameterize: simplify tests

raimis:
- [x] Test plot data
   - [x] Enable plot data writing to files
   - [x] Store the files in the test data
- [x] Improve error messages
- [x] Clean up


htmd/parameterization/dihedral.py
```diff
@@ -410,6 +410,14 @@ def plotDihedralEnergies(self, idihed):
         fitted_energy = self._fitted_energies[idihed] - np.min(self._fitted_energies[idihed])
         indices = np.argsort(angle)
 
+        path =  os.path.join(self.result_directory, self._names[idihed])
```
j3mdamas: I would like to keep the `.svg` inside the join function

htmd/parameterization/dihedral.py
```diff
@@ -410,6 +410,14 @@ def plotDihedralEnergies(self, idihed):
...
+
+        if write_data:
+            header = 'angle  QM_ref   MM_init  MM_fit'
```
j3mdamas: Since we are at it, you could make the header follow the `fmt` of the `savetxt` (fixing the size in general, and changing to f or n, depending on the writing)

htmd/parameterization/dihedral.py
raimis: Why?

htmd/parameterization/dihedral.py
raimis: Don't get your idea?

htmd/parameterization/dihedral.py
j3mdamas: for consistency purposes. adding `.svg` after path building through join may create erroneous strings

htmd/parameterization/dihedral.py
j3mdamas: just did a commit with my idea

raimis:
@j3mdamas is this finished?

j3mdamas:
As soon as it passes, let's merge

PR #569, opened 2018-02-22
Init SmallMol refactoring

cuzzo87:
- [x] deal with erroneous inputs
    - [x] Moved the input parsing  from the `__init__` into a function.
    - [x] Added a functionality to load erroneous pdb/mol2 file by converting them to 'sdf' (InputToOutput in util.py)
- [ ] wrap rdkit functionalities:
    - [x] get if molecule is chiral. Can return the stereocenters and the chiral type ('R', 'S')  
    - [x] get atom neighbors (from id or atomtype, filters baed on atomtype and bondtype)
    - [ ] get atom properties
         - [ ] hybridizations
         - [ ] valence
         - [x] formalcharge
         - [x] charge
    - [x] get_element
    - [x] get_atoms
    - [x] get_atom
    - [x] set atom properties
    - [x] set_name (set the molecule name in rdkit Molecule. If None is set when created is set as 'UNK')   
  
- [ ] additional
   - [x] set and get molecule properties
   - [x] set and get atom properties
   - [x] list molecule and atoms properties  
   - [x] copy 
   - [x] get_mol to return rdkit Molecule object
   - [x] improve conformer generators/ get conformers etc
   - [x] write onformers as seperate method  from generate one and improvements
- [ ] depiction 
   - [x] 2D
   - [x] 3D 
   - [x] optimization (valid only for 3D)
   - [x] highlighting atoms with colors
   - [x] atomlabels
   - [x] multiple mols into a grid
- [ ] SmallMolStack:
   - [ ] compound similarity
   - [ ] substruct match
   - [ ] max common structure
   - [ ] Not druggable compounds: Lipinski and Oprea rules; PAINS; REOS


stefdoerr:
Hm I don't like much the idea of storing properties of the molecule in numpy arrays in it. If you go and modify the rdkit object you deprecate the values in the numpy arrays. I think we should sit down and discuss the class design and plans for the class. Otherwise you are developing blindly and wasting time I think

cuzzo87:
Ok. This afternoon if you can we can discuss it. The reason for changing this behaviour was to speed up the molecule loading for big databases and also because the properties are not safely saved (rdkit c++ code not safely pickle these objects)

cuzzo87:
Hi,

I'm going to close this PR, squash the commits into one, and do a new one. It will be easier to review.

Alberto and Joao

PR #574, opened 2018-02-28
Rel 1.12.x intermediary merging into devel

j3mdamas:
just need to do this to release htmd-data


j3mdamas:
@stefdoerr sorry it was a mistake, I don't think this needs review, just pass travis

PR #581, opened 2018-03-07
Small fixes on tests to pass on local deployments (acemanager install…

j3mdamas:
…ations)

Very easy to review


stefdoerr:
You mixed two commits. But ok I guess

j3mdamas:
the detect.py was causing to fail tests if there was a configuration file, because the imports on doctests were giving output. so it's part of the same problem

j3mdamas:
so I "force" the import before the doctests

PR #582, opened 2018-03-08
Toni fix ionize standard names

### LINKED ISSUE #463: Accept standard ion names in ionize
I think ionize() can be made to accept standard PDB names instead of ff-specific ones, so it's one less thing to worry about when switching ff. I think the builders already have all the aliases in place.

j3mdamas:
I've cleaned and tested Toni's changes. Should close #463


j3mdamas:
Can I merge  @stefdoerr? :)

j3mdamas:
I didn't and you're right, I should. Will do before merging.

j3mdamas:
Just did, simulation works, ions look fine. Merging.

PR #591, opened 2018-03-15
Fix nglviewdeps

### LINKED ISSUE #584: NGLview is not showing structures
I still don't understand why, probably a browser problem. I've opened this issue https://github.com/arose/nglview/issues/733.

I'll try to see if I can get more information, but it needs a solution before 1.12.x is released.

j3mdamas:
Will close #584

### DIFF
--- a/package/htmd-deps/write_meta_yaml.py
+++ b/package/htmd-deps/write_meta_yaml.py
@@ -25,18 +25,26 @@ requirements:
 deps = []
 with open(os.path.join(workdir, 'DEPENDENCIES'), 'r') as f:
     for line in f:
-        deps.append(line.strip().lower())
+        # Remove comments and split
+        packagematch = line.split('#')[0].split()
+        # Append dict of the conda package match specification
+        if packagematch:
+            deps.append(dict(zip(['name', 'version', 'build_string'], packagematch)))
 
 # Get all installed packages from conda
 packages = check_output(['conda', 'list', '--json']).decode('utf8')
 packages = json.loads(packages)
 
 # Find the version of each dependency and add them to the meta.yaml file
-found = {key: False for key in deps}
+found = {dep['name']: False for dep in deps}
 for p in packages:
     name = p['name'].lower()
-    if name in deps:
-        text += '    - {} >={}\n'.format(name, p['version'])
+    if name in [dep['name'] for dep in deps]:
+        dep = next(dep for dep in deps if dep['name'] == name)
+        if 'version' not in dep:
+            text += '    - {} >={}\n'.format(name, p['version'])
+        else:
+            text += '    - {} {}\n'.format(name, dep['version'])
         found[name] = True
 
 # Check if all dependencies were found. If not print which and exit with error
@@ -46,5 +54,6 @@ if not all(found):
             print('Could not find dependency "{}". Exiting with error.'.format(key))
     sys.exit(1)
 
+# Write htmd-deps meta.yaml file
 with open(os.path.join(workdir, 'meta.yaml'), 'w') as f:
     f.write(text)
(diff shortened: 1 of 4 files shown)


package/htmd-deps/write_meta_yaml.py
```diff
@@ -25,18 +25,26 @@
 deps = []
 with open(os.path.join(workdir, 'DEPENDENCIES'), 'r') as f:
     for line in f:
-        deps.append(line.strip().lower())
+        # Remove comments and split
+        packagematch = line.split('#')[0].split()
+        # Append dict of the conda package match specification
+        if packagematch:
+            deps.append(dict(zip(['name', 'version', 'build_string'], packagematch)))
```
stefdoerr: Doesn´t this assume that the split will return 3 values? I think zip will fail on lists of different lengths no?

package/htmd-deps/write_meta_yaml.py
```diff
@@ -25,18 +25,26 @@
...
 packages = json.loads(packages)
 
 # Find the version of each dependency and add them to the meta.yaml file
-found = {key: False for key in deps}
+found = {dep['name']: False for dep in deps}
 for p in packages:
     name = p['name'].lower()
-    if name in deps:
-        text += '    - {} >={}\n'.format(name, p['version'])
+    if name in [dep['name'] for dep in deps]:
```
stefdoerr: You could do `if name in list(found.keys())` no? To not redo the list/dict comprehension

stefdoerr: Or just `if name in found` actually I think works

sync_acellera_conda_channel_deps.py
```diff
@@ -30,14 +31,22 @@
...
-    ['omnia', 'nglview'],  # Adding both omnia and bioconda nglview because they provide different builds
     ['rdkit', 'rdkit'],
     ['rdkit', 'boost'],
     ['conda-forge', 'nlopt'],
     ['conda-forge', 'periodictable']
 ]
 
+# Package arguments
+
+allpackages = [i[1] for i in packages]
```
stefdoerr: Mmm, it's a style thingy, sorry if being pedantic but use `i` for loop indexes only. Use `p` or `package` here better.

package/htmd-deps/write_meta_yaml.py
j3mdamas: It surprisingly works, I tested it (and it passes Travis)! But if you think it's unsafe, please let me know how to do it.

sync_acellera_conda_channel_deps.py
j3mdamas: totally agree. I was going to use package, but it's really a list with channel and package. I'll use p.

package/htmd-deps/write_meta_yaml.py
j3mdamas: using `if name in found` here is confusing, I think. I've made a new proposal to avoid redoing list comprehension, I hope you approve :)

stefdoerr:
Ah there is the zip thing before merging. I'll give it a try tomorrow, see how python behaves

j3mdamas:
Please do. In my tests, I found zero problems, but they weren't thorough
maybe.

package/htmd-deps/write_meta_yaml.py
stefdoerr: You are right. It apparently works
```
>>> list(zip(['name', 'value', 'index'], ['free', 'test']))
[('name', 'free'), ('value', 'test')]
```
Personally I don't like this behaviour of python but that's irrelevant to the PR.

stefdoerr:
Feel free to merge

PR #596, opened 2018-03-19
fftype needs to be consistent. if it returns mol it ought to be a copy

stefdoerr:
Right now it's modifying it in place and returning it which is very confusing. Since we have decided a while ago to not modify objects in-place inside functions (and methods only modify in place and don't return copies) this should also be consistent. I am sure I'll find a few more lying around at later points.


raimis:
@stefdoerr I need to refactor `fftype` in general at some point. If the tests passes, I'm fine with this change.

PR #597, opened 2018-03-19
add assertion and dtype conversions ffevaluate

stefdoerr:
Dominic found a really nasty bug with wrong box dimensions where numba was reading garbage from out-of-bounds memory. Added assertions and dtype conversions to avoid future problems.

### DIFF
--- a/htmd/ffevaluation/ffevaluate.py
+++ b/htmd/ffevaluation/ffevaluate.py
@@ -34,6 +34,8 @@ class FFEvaluate:
             box = np.zeros((3, coords.shape[2]), dtype=np.float32)
         # from IPython.core.debugger import set_trace
         # set_trace()
+        assert box.shape[0] == 3 and box.shape[1] == coords.shape[2], 'Box dimensions have to be (3, numFrames), your Molecule has box of shape {}'.format(box.shape)
+
         energies, forces, atmnrg = _ffevaluate(coords.astype(np.float32), box.astype(np.float32), *self._args)
         return _formatEnergies(energies[:, 0].squeeze())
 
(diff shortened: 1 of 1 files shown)


htmd/ffevaluation/ffevaluate.py
```diff
@@ -34,6 +34,8 @@ def run(self, coords, box=None):
             box = np.zeros((3, coords.shape[2]), dtype=np.float32)
         # from IPython.core.debugger import set_trace
         # set_trace()
+        assert box.shape[0] == 3 and box.shape[1] == coords.shape[2], 'Box dimensions have to be (3, numFrames), your Molecule has box of shape {}'.format(box.shape)
+
```
raimis: This should throw `ValueError` exception, rather than assert. As the error can be triggered by a user directly passing a wrong `box`.

htmd/ffevaluation/ffevaluate.py
```diff
@@ -256,9 +258,11 @@ def ffevaluate(mol, prm, betweensets=None, cutoff=0, rfa=False, solventDielectri
     >>> prm = parmed.amber.AmberParameterSet.from_structure(prmtop)
     >>> energies, forces, atmnrg = ffevaluate(mol, prm, betweensets=('resname SOD', 'water'))
     """
+    assert mol.box.shape[0] == 3 and mol.box.shape[1] == mol.coords.shape[2], 'Box dimensions have to be (3, numFrames), your Molecule has box of shape {}'.format(mol.box.shape)
+
```
raimis: This should throw `ValueError` exception, rather than assert. As the error can be triggered by a user directly passing a wrong `box`.

stefdoerr:
it's a conda timeout. I'll merge anyway since it passed 3/4

stefdoerr:
Ah damn a few seconds too late. But ok I can change it

raimis:
@stefdoerr if you ask for a review, then wait for the review, please.

j3mdamas:
pulled the merge trigger too fast :P

stefdoerr:
It was a "either/or" review request but ok I'll only put single persons
then. Made a new PR with your fix anyway

PR #599, opened 2018-03-19
bugfix for fftype. Masses are usually correct length but zeros

stefdoerr:
The check to substitute masses in fftype was wrong. Fixed it now


raimis:
I have a feeling with won't pass the tests... :D

stefdoerr:
Hmmm we might get away with it since `mol2` files don't contain masses :smile:

stefdoerr:
That was wishful thinking :P But ok it's just cgenff tests not passing due to masses in rtf. I'll fix it quick

stefdoerr:
@raimis it passes. Are you ok with a merge?

PR #611, opened 2018-03-22
SmallMol refactoring

cuzzo87:
The first step of refactoring. 
Missing another final step of cleaning and docstrings.

Finally, I will create a unittest for all the functionalities


j3mdamas:
@stefdoerr, I've helped Alberto squashing everything into one single commit. I think we can start reviewing this.

htmd/smallmol/chemlab/periodictable.py
```diff
@@ -0,0 +1,189 @@
...
+            The number of atoms that miss for completing the atom valence
+
+        """
+        normbondvalence = sum([ 1.5 if int(b) == 12 else int(b) for b in btypes  ])
+
+        missatoms = int(valence - normbondvalence)
+
+        return missatoms
+
+    def getRadiusBond(self, element):
```
stefdoerr: getBondedRadius sounds clearer. I thought this returned the radius of the bond whatever that would be :P

htmd/smallmol/chemlab/periodictable.py
```diff
@@ -0,0 +1,189 @@
...
+
+        pT = self.PeriodicaTable
+        bradius = pT.GetRb0(element)
+
+        return bradius
+
+
+    def getMissingValence(self, atomidx, smallmol, formalcharge=0):
+        """
+        Returns the number of missing atom to complete the atom valence based on the types of bonds and formalcharge
```
stefdoerr: It's number of missing electrons no?

htmd/smallmol/chemlab/periodictable.py
```diff
@@ -0,0 +1,189 @@
...
+            The index of the atom
+        smallmol: htmd.smallmol.smallmol.SmallMol
+            The SmallMol object
+        formalcharge: int
+            The formalcharge of the atom
+            Default: 0
+
+        Returns
+        -------
+        missingatoms: int
```
stefdoerr: Same here, missing electrons?

htmd/smallmol/chemlab/periodictable.py
```diff
@@ -0,0 +1,189 @@
...
+        # Some atoms can have multiple valid valences (like S, P). For them I should predict the most reliable one.
+        # At the moment not implemented. I will raise an error now.
+        if len(valences) == 1:
+            valence = valences[0]
+        else:
+            class Notmplemented(Exception):
+                pass
+            raise Notmplemented('The missing valence is not implemented for atom with more than one valid valence')
+
+        missingatoms = self._evaluateMissingAtoms(valence,  btypes)
```
stefdoerr: evaluateMissingElectrons no?

htmd/smallmol/chemlab/builder.py
```diff
@@ -0,0 +1,406 @@
...
+
+    def addHydrogens(self, onlyExplicit=False):
+        """
+        Adds the hydrogens to fill up the valence of the atoms. With onlyExplicit is possible to add only the polar ones
+
+        Parameters
+        ----------
+        onlyExplicit: bool
+            If True, only the polar hydrogens will be added
+            Default: False
```
stefdoerr: Small detail but there is no need to document the default values as they can be seen in the function signature. So better avoid documenting them.

htmd/smallmol/chemlab/builder.py
```diff
@@ -0,0 +1,406 @@
...
+        pT = self.periodictable
+        sm = self.smallmol
+
+        # polar hydrogens are the ones bonded to hetero atoms (no carbon)
+        if onlyExplicit:
+            heavy_sel = ' '.join(_hetero_atoms)
+        else:
+            heavy_sel = ' '.join(_heavy_atoms)
+
+        atomidxs = sm.get( 'idx', 'element {}'.format(heavy_sel))
```
stefdoerr: If you end up calling this function many times, atomselections can become a bottleneck. I tend to work my way around them. For example, you could achieve the same selection with `np.in1d(sm.element, _heavy_atoms)` which returns a boolean mask with True if element is in _heavy_atoms

stefdoerr: Ah although this is the `SmallMol.get` so I guess you don't use atomselection. Ignore my comment. Sorry

htmd/smallmol/chemlab/builder.py
```diff
@@ -0,0 +1,406 @@
...
+            for i in range(n_h):
+                new_atom = pT.getAtom('H', a)
+                h_atoms.append(new_atom)
+
+        self._addAtoms(h_atoms)
+
+
+    def removeHydrogens(self, removePolars=True, removeNonPolars=True):
+        """
+        Removes the hydrogens. Is it possible to choose the type of hydrogens to remove: polar and not polar
```
stefdoerr: `Is it possible` = `It is possible`

htmd/smallmol/chemlab/builder.py
```diff
@@ -0,0 +1,406 @@
...
+
+        neighborsIdx = np.concatenate(sm.neighbors[hydrogensIdx])
+        neighbors = sm.element[neighborsIdx]
+        hydrogensPolarIdx = hydrogensIdx[np.where( neighbors != 'C' )]
+        hydrogensNonPolarIdx = hydrogensIdx[np.where(neighbors == 'C')]
+
+
+        hydrogensIdx = np.array([],dtype=np.int)
+        if removePolars:
+            hydrogensIdx = np.append( hydrogensIdx, hydrogensPolarIdx)
```
stefdoerr: Don't append to numpy arrays (it's slow). Use lists and at the last step convert to np.array.

Also make sure you use the correct int dtype. Do you want `int32` or `int64` or even more correct `uint32`? Indexes cannot be negative and I doubt you will have more than 4billion atoms so I guess `uint32` is enough.

```python
hydrogensIdx = []
if bla:
   hydrogensIdx.append(x)
if blu:
   hydrogensIdx.append(y)
hydrogensIdx = np.array(hydrogensIdx , dtype=np.int32)
```

htmd/smallmol/chemlab/builder.py
```diff
@@ -0,0 +1,406 @@
...
+            List of the atom data to add
+        """
+
+        sm = self.smallmol
+
+        #c = 1
+        for a in atoms:
+            n_atom = sm.numAtoms
+           # print('placing {} to {}'.format(n_atom, a['attachTo']))
+            sm.__dict__['idx'] = np.append(sm.idx, n_atom)
```
stefdoerr: There is no need to use `__dict__` here. Just do `sm.idx` etc.

stefdoerr: Here I guess the append is ok since you would need to convert to a list first and back again.

htmd/smallmol/chemlab/builder.py
```diff
@@ -0,0 +1,406 @@
...
+            nbrVect = nbr_coords - heavy_coords
+
+            nbrVect = _normalizeVector(nbrVect)
+            nbrVect = -nbrVect
+
+            perpVect = _getPerpendicular(nbrVect)
+            perpVect = _normalizeVector(perpVect)
+
+            if heavy_hybridization == 4:
+                rot_matrix = _getRotationMatrix(perpVect, (180-109.471), deg=True)
```
stefdoerr: `180-109.471` Document such freakish numbers :P Maybe it makes sense to a chemist but I have no clue if this is a bug haha.

htmd/smallmol/chemlab/builder.py
```diff
@@ -0,0 +1,406 @@
...
+            #TODO based on chiral
+            sm.__dict__['_chiraltags'] = np.append(sm._chiraltags, 0)
+            # Development note: removed assignement explicitHs
+
+            # coords
+            sm.__dict__['coords'][n_atom]  = self._getAtomCoords(n_atom, a['attachTo'])
+
+            #c += 1
+
+    def _getAtomCoords(self, atomIdx, attachToIdx):
```
stefdoerr: Maybe add VSEPR to the function name. Otherwise I would expect it to just return atom coordinates

htmd/smallmol/chemlab/builder.py
```diff
@@ -0,0 +1,406 @@
...
+        for n_set, b_set in zip(sm.__dict__['neighbors'], sm.__dict__['bondtypes']):
+            new_n_set = []
+            new_b_set = []
+            for n, b in zip(n_set, b_set):
+                if n not in ids:
+                    new_n_set.append(n)
+                    new_b_set.append(b)
+            new_neighbors.append(new_n_set)
+            new_bondstype.append(new_b_set)
+        sm.__dict__['neighbors'] = np.array(new_neighbors)
```
stefdoerr: No need for `__dict__` anywhere in this function except in the first loop.

htmd/smallmol/smallmol.py
```diff
@@ -1,105 +1,654 @@
...
+
+    mol, ignore_errors=False, force_reading=False, fixHs=True, removeHs=False
+    Parameters
+    ----------
+    mol: rdkit.Chem.rdchem.Mol  or filename or smile or htmd.smallmol.smallmol.SmallMol
+        (i) Rdkit molecule or (ii) Location of molecule file (".pdb"/".mol2") or (iii) a smile string or iv) another
+        SmallMol object
+    ignore_errors: bool
+        If True, errors will not be raised.
+        Default: False
```
stefdoerr: Don't document defaults

htmd/smallmol/smallmol.py
```diff
@@ -1,105 +1,654 @@
...
+
+    _mol_dims = {'_mol': (0,),
+                 'ligname': (0,),
+                 'totalcharge': (0,)
+                   }
+
+    def __init__(self, mol, ignore_errors=False, force_reading=False, fixHs=True, removeHs=False):
+
+
+        self.mol_fields = self._mol_fields.copy()
```
stefdoerr: No need to copy class static variables unless you plan on modifying them (which I don't see you doing anywhere). If you want to make them public just remove the underscore.

htmd/smallmol/smallmol.py
```diff
@@ -1,105 +1,654 @@
...
+        """
+        Returns the number of conformers in the rdkit molecule object
+
+        Returns
+        -------
+        nconfs: int
+            The number of conformers in the molecule
+        """
+        # getting the number of conformers from the shape of coords array
+        _nConformers = self.coords.shape[-1]
```
stefdoerr: detail: I'd prefer it returning `self.coords.shape[2]`. That way if for some reason you messed up and `self.coords` is two dimensional it will crash instead of returning `3` for `xyz`

htmd/smallmol/smallmol.py
```diff
@@ -1,105 +1,654 @@
...
+        """
+        Create a copy of the molecule object
+
+        Returns
+        -------
+        newsmallmol : :class:`SmallMol`
+            A copy of the object
+        """
+        newsmallmol = deepcopy(self)
+        return newsmallmol
```
stefdoerr: By the way, have you ever checked if this copies the `_mol` object? I am a bit worried about `rdKit` objects.

htmd/smallmol/smallmol.py
```diff
@@ -1,105 +1,654 @@
...
+            Default: None
+
+        Example
+        -------
+        >>> sm.setPropsAtoms('myTag', ['a', 'b', 'a', 'c'], aIdxs=[0,2,5,3]) # doctest: +SKIP
+        >>> sm.myTag
+        array(['a', None, 'b', 'c', None, 'a', None, None, None, None, None, None,
+       None, None, None, None, None, None], dtype=object)
+        """
+        class ValuesLengthErrors(Exception):
```
stefdoerr: rename to `ValueLengthError`. No plural needed if you see how `ValueError` is called.

htmd/smallmol/smallmol.py
```diff
@@ -228,118 +814,783 @@ def get_voxels(self, center=None, size=24, resolution=1., rotation=None,
...
+            The list of conformer id to delete. If None, all are removed except the first one
+            Default: None
+        """
+
+        _nConformers = self.numConformers
+
+        if ids is None:
+            ids = np.arange(_nConformers)
+        elif not isinstance(ids, list):
+            raise TypeError("The ids argument should be list of confermer ids")
```
stefdoerr: `confermer` = `conformer`

htmd/smallmol/chemlab/periodictable.py
```diff
@@ -0,0 +1,189 @@
...
+
+        """
+
+        pT = self.PeriodicaTable
+        bradius = pT.GetRb0(element)
+
+        return bradius
+
+
+    def getMissingValence(self, atomidx, smallmol, formalcharge=0):
```
stefdoerr: I would switch the order of arguments to be `smallmol, atomidx` Because when you use the formal charge it looks messy having the object in the middle of the properties.

htmd/smallmol/tools/clustering.py
```diff
@@ -0,0 +1,423 @@
...
+    print('{} problematic molecules. Indexes: {}'.format(len(wrong), wrong))
+
+    clustmethod = getattr(this_module, '_{}Clustering'.format(method))
+
+    if method not in ['shape', 'mcs']:
+        matrix = clustmethod(rdkitMols_list)
+
+    else:
+        aprun = ParallelExecutor(n_jobs=-1)  # _config['ncpus'])
+        matrix = aprun(total=len(rdkitMols_list), desc='{} Distance'.format(method)) \
```
stefdoerr: You are parallelizing over parallelizations, since your methods also parallelize! Are you really sure you want to do this?

htmd/smallmol/util.py
```diff
@@ -71,4 +181,227 @@ def drawIsoSurface(values3d, resolution=1., plot_center=None, viewer=None):
...
+    print('Alignment with a RMSD of {}'.format(rmsd))
+    coords_new = sm_rdkit.GetConformer().GetPositions()
+
+    sm_new = SmallMol(smallmol, fixHs=False)
+    sm_new.removeConformers()
+    sm_new.coords = coords_new[:, :, np.newaxis]
+
+    return sm_new
+
+def inputToOutput(input_file, input_format, output_format):
```
stefdoerr: rename this to `openbabelConvert` or `convertFileType`. `inputToOutput` is a too generic name and makes no sense out of context

stefdoerr:
Ok I'm done with the PR. Take a look at my comments @cuzzo87 and better in the `Files changed` panel because here you cannot see all of them. Take care to also expand the collapsed diffs in the `Files` panel as sometimes `github` decides to hide stuff...

htmd/smallmol/chemlab/periodictable.py
cuzzo87: done it

htmd/smallmol/chemlab/builder.py
cuzzo87: Always random typing :).  Done it.

htmd/smallmol/smallmol.py
cuzzo87: done it.

htmd/smallmol/chemlab/builder.py
cuzzo87: removed all the documented default value in the whole smallmol module

htmd/smallmol/chemlab/builder.py
cuzzo87: I see. Due to the fact that I am using two numpy (in your example x and y are numpy arrays). I did like that:
```
hydrogensIdx = []
if removePolars:
   hydrogensIdx.extend(hydrogensPolarIdx.tolist())
if removeNonPolars:
   hydrogensIdx.extend(hydrogensNonPolarIdx.tolist())
hydrogensIdx = np.array(hydrogensIdx, dtype=np.uint32)
```

htmd/smallmol/chemlab/builder.py
cuzzo87: done it. Should I have to do like that in all the parts where I used self.__dict__ ?

htmd/smallmol/chemlab/builder.py
cuzzo87: Actually the method is returning the coordinates of that atom. Maybe should be called _computeCoord. I wait your opinion on that

htmd/smallmol/chemlab/builder.py
cuzzo87: This comes from the tetrahedral geometry of the VSEPR theory. SP3 has tetrahedral geometry with angles of 109.471. I have just kept as 180 - 109.471 to be clear for a chemist.

htmd/smallmol/chemlab/builder.py
cuzzo87: done it.

htmd/smallmol/chemlab/periodictable.py
cuzzo87: I see your point. Modified as you suggested

htmd/smallmol/chemlab/periodictable.py
cuzzo87: Valence is the number of external electrons that can technically be involved in bonds.  With this method I get the number of atoms expected to be bonded to the atom based on valence and formalcharge. I would keep the name I set. do you agree?

htmd/smallmol/chemlab/periodictable.py
cuzzo87: same as before

htmd/smallmol/chemlab/periodictable.py
cuzzo87: Exactly as before. I know the valence and checking how many atoms are present at the moment I get how many are missing

htmd/smallmol/smallmol.py
cuzzo87: done it

htmd/smallmol/smallmol.py
cuzzo87: _chiraltags is modified in chirals compounds. Due to the fact that is not a chemist readable value I decided to hide it with the underscore. I mean that is necessary to copy it because it would be modified, but I also would like to keep it hidden. If you have a better way to do it I will modify it.

Regarding the '_mol'. This is the origianl rdkit mol. I keep it for debugging purpose only. Here the reason why it is hidden with the underscore. Can I keep it like that or do you suggest a better way?

htmd/smallmol/smallmol.py
cuzzo87: I see. thanks for the suggestion. Done it

htmd/smallmol/smallmol.py
cuzzo87: The rdkit mol is copied correctly. The properties stored in the molecule and atoms are probably lost. All the important infomration are already stored in the __dict__, thus loosing that info are not anymore a problem.  Moreover, as I commented above, the '_mol' is just for debugging purpose

htmd/smallmol/smallmol.py
cuzzo87: done it

htmd/smallmol/util.py
cuzzo87: you are completely right. Modified it

cuzzo87:
For my side, I modified accordingly to what Stefan suggested. Few stuffs are still pending, but I think it can be already merged. What do you think @stefdoerr

j3mdamas:
There was a small conflict which I resolved. I'd like to do a quick review by this week.

htmd/smallmol/chemlab/builder.py
stefdoerr: Or you can append np.arrays to a list and at the end `np.vstack/hstack/stack` them. I usually do that. No converting to list and back to array would be ugly :D So either the way it was before with the correct dtype or as I said with stacking :)

htmd/smallmol/chemlab/builder.py
stefdoerr: `self.__dict__` should only be used when you don't know the name of the attribute. I.e. 
```python
x = 'names'
x = 'resid'
self.__dict__[x]
```
If you know you want the resid just do directly `self.resid`. 

Practically to be more correct, I was schooled recently by @Acellera/david that using the `__dict__` is not the best way. There are in python two functions called `getattr` and `setattr` which work as `setattr(sm, x, value)` and are probably the most correct way of doing it. I will try using that from now on but I don't want to make you change all that now. Too much work and it's my fault for doing this the wrong way to begin with

htmd/smallmol/chemlab/builder.py
stefdoerr: Hm yes, compute sounds better if you do such have computation. `_computeAtomCoords` maybe?

htmd/smallmol/smallmol.py
stefdoerr: But `_mol_fields = ['ligname', 'totalcharge', '_mol']`. This is a list of strings. You are copying a list of strings. Do you mean that you append to this list of strings somewhere a new string?

stefdoerr:
@cuzzo87 look at my few new comments. Thanks a lot for all the changes (and sorry for making you do so many)!

htmd/smallmol/smallmol.py
cuzzo87: Yes. This list of strings was created to store any kind of molecule property. For instance, the SmallMolLib use this list to store the sdf fields.

htmd/smallmol/chemlab/periodictable.py
cuzzo87: Technically you are right.  I did not consider the possibility to directly add a double or triple bond. The current version of the method was designed to support the addition and deletion of hydrogens only. 
For the future (imagine you want to add a  C=O), you initially connect O to C with a single C-O; then you try to convert it to a double.

htmd/smallmol/chemlab/builder.py
cuzzo87: Hoping is correct, I modified with ```np.hstack``` like that:
```
hydrogensIdx = np.array([], dtype=int)
if removePolars:
   hydrogensIdx = np.hstack((hydrogensIdx, hydrogensPolarIdx))
if removeNonPolars:
   hydrogensIdx = np.hstack((hydrogensIdx, hydrogensNonPolarIdx))
```

htmd/smallmol/chemlab/builder.py
cuzzo87: I will keep it in mind and in the next future I'll refactor this by considering ```setattr``` and ```getattr```.

htmd/smallmol/smallmol.py
stefdoerr: I see okay thanks! :)

htmd/smallmol/chemlab/builder.py
stefdoerr: Wrong dtype though :D As I said `uint32` is more than enough. It cannot go negative so `int` which is `int64` is overkill (sorryyyyy)

htmd/smallmol/chemlab/periodictable.py
stefdoerr: So woudln't it be better to not say missing atoms but say missing bonds?

stefdoerr:
I also don't want to drag this PR to infinity either hahaha

PR #612, opened 2018-03-22
Parameterize: fix logger configuration

raimis:
Parameterize logger message are silenced, if `htmd.ui` is not imported, which, as a side effect, configures loggers.

### DIFF
--- a/htmd/parameterization/cli.py
+++ b/htmd/parameterization/cli.py
@@ -9,6 +9,7 @@ import argparse
 import logging
 import numpy as np
 
+import htmd.ui # This is needed to configure logging
 from htmd.version import version
 from htmd.queues.localqueue import LocalCPUQueue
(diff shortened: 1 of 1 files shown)


j3mdamas:
@stefdoerr should we make it more general and put logging on `htmd.__init__`?

raimis:
@stefdoerr maybe I don't get how this logger configuration should work. Is any better solution for this?

j3mdamas:
Yeah, my proposal here is to move:

```
try:
    logging.config.fileConfig(os.path.join(htmd.home.home(), 'logging.ini'), disable_existing_loggers=False)
except:
    print("HTMD: Logging setup failed")
```

from `htmd.ui` into `htmd.__init__`

stefdoerr:
Yeah I think @j3mdamas is right.

j3mdamas:
Let me close this and open a new fix, to include the release version.

PR #615, opened 2018-03-23
Implements versioning on LSF (9 and 10)

### LINKED ISSUE #603: Upgrade LsfQueue to use LSF version 10
Keep backwards compatibility with LSF version 9

Main issue is to accommodate the new `-gpu` flag with the old settings for gpus

Deployments that work with LSF 10 are broken if they use the -gpu flag.

j3mdamas:
Updated ngpus to work with the new version and implements the new gpu_options
Updates the docstring
Depends on Acellera/protocolinterface#5 and closes #603 
@stefdoerr you need to review Acellera/protocolinterface#5 first :)

### DIFF
--- a/htmd/queues/lsfqueue.py
+++ b/htmd/queues/lsfqueue.py
@@ -21,31 +21,40 @@ class LsfQueue(SimQueue, ProtocolInterface):
 
     Parameters
     ----------
+    version : [9, 10], int, default=9
+        LSF major version
     jobname : str, default=None
         Job name (identifier)
-    queue : str or list of str, default=None
-        The queue or list of queues to run on. If list, it attempts to submit the job to the first queue listed. If that
-        queue cannot be used, the next queue listed is considered and this process is repeated.
+    queue : list, default=None
+        The queue or list of queues to run on. If list, it attempts to submit the job to the first queue listed
     app : str, default=None
         The application profile
     ngpu : int, default=1
         Number of GPUs to use for a single job
+    gpu_options : dict, default=None
+        Number of GPUs to use for a single job(dict format: {['mode': ['shared', 'exclusive_process'],]['mps': ['yes',
+        'no'],]['j_exclusive': ['yes', 'no']]})
     ncpu : int, default=1
         Number of CPUs to use for a single job
     memory : int, default=4000
         Amount of memory per job (MiB)
     walltime : int, default=None
         Job timeout (hour:min or min)
+    resources : list, default=None
+        Resources of the queue
+    outputstream : str, default='lsf.%J.out'
+        Output stream.
+    errorstream : str, default='lsf.%J.err'
+        Error stream.
+    datadir : str, default=None
+        The path in which to store completed trajectories.
+    trajext : str, default='xtc'
+        Extension of trajectory files. This is needed to copy them to datadir.
     envvars : str, default='ACEMD_HOME'
         Envvars to propagate from submission node to the running node (comma-separated)
-    prerun : list of strings, default=None
+    prerun : list, default=None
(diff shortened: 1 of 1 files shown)


htmd/queues/lsfqueue.py
```diff
@@ -21,31 +21,40 @@ class LsfQueue(SimQueue, ProtocolInterface):
...
-        queue cannot be used, the next queue listed is considered and this process is repeated.
+    queue : list, default=None
+        The queue or list of queues to run on. If list, it attempts to submit the job to the first queue listed
     app : str, default=None
         The application profile
     ngpu : int, default=1
         Number of GPUs to use for a single job
+    gpu_options : dict, default=None
+        Number of GPUs to use for a single job(dict format: {['mode': ['shared', 'exclusive_process'],]['mps': ['yes',
+        'no'],]['j_exclusive': ['yes', 'no']]})
```
stefdoerr: Hm, no sorry for not mentioning this earlier I didn't have time :( This is very confusing for me. I understand the idea but the `[ ]` just make me think of lists in python, so I'd try to create a dict of lists. Maybe just point to the LSF docs for available options and say to put them in a dict? Otherwise we could make a "class" for it but I'd prefer not

htmd/queues/lsfqueue.py
```diff
@@ -21,31 +21,40 @@ class LsfQueue(SimQueue, ProtocolInterface):
...
         Error stream.
+    datadir : str, default=None
+        The path in which to store completed trajectories.
+    trajext : str, default='xtc'
+        Extension of trajectory files. This is needed to copy them to datadir.
+    envvars : str, default='ACEMD_HOME'
+        Envvars to propagate from submission node to the running node (comma-separated)
+    prerun : list, default=None
+        Shell commands to execute on the running node before the job (e.g. loading modules)
+
```
stefdoerr: FYI `ProtocolInterface` keeps track of the order with which you declare the parameters. So if you want to switch the order in the documentation just declare them in that order in the `__init__` ;)

j3mdamas: this reordering was just because I created the docstring manually last time instead of generating it with the _printDocString you showed me today

stefdoerr: yes I just mentioned in in case you prefered the old order

htmd/queues/lsfqueue.py
j3mdamas: well, the LSF docs will not explain the dictionary format.

The only solution I see for this right now is split the `gpu_options` into three string arguments.

Or wait, what about this forma?:
```
(valid dict entries: {'mode': <'shared' or 'exclusive_process'>}, {'mps': <'yes' or 'no'>},
 {'j_exclusive': <'yes' or 'no'>})
```

htmd/queues/lsfqueue.py
j3mdamas: I'm actually OK with the order being generated from the `_printDocString` :)

htmd/queues/lsfqueue.py
stefdoerr: yeah reads slightly better for me

j3mdamas: fixed

j3mdamas:
https://github.com/Acellera/protocolinterface/pull/5 was merged and this was updated according to it.

I think all is solved and it can be merged.

j3mdamas:
Sorry, the default behaviour of LSF 10 was talked with Matt but I had not changed it. Now it's complete.

j3mdamas:
Done and passed tests. I addressed all review comments. Is it OK, @stefdoerr?

stefdoerr:
Yesh :smiley:

PR #618, opened 2018-03-28
Adding unique residue and atom identifiers to get rid of renumberResidues

stefdoerr:
This is needed for fixing issue #537 
I just create a way to store information about an atom that identifies it uniquely to then be able to get the same atom after possibly changing atom order or number of atoms. This is necessary for example for defining disulfide bridges in the builders.


stefdoerr:
To clarify, these classes are for the moment not meant to be used by users. Just used internally by the builders (as you will see in the next PR once this is merged)

stefdoerr:
I now pushed the rest of the PR as well as we agreed with @j3mdamas . This PR should make `renumberResidues` obsolete and thus solve also the issue with using it together with `autoSegment1`. Practically what was needed was a unique way to identify atoms and residues because before I was doing it differently in different parts of HTMD. Now that it's standardized it also takes into account residue insertions and cleans up the whole mess.

j3mdamas:
Let me merge master into this before reviewing

j3mdamas:
I'm a bit concerned that we cannot include this fix into stable, but I really believe that this is only feasible on master now.

stefdoerr:
It's not extremely urgent I think as it can be worked around by doing
renumberResidues after autoSegment. We can just make another stable release
sooner

htmd/builder/amber.py
```diff
@@ -325,21 +325,37 @@ def build(mol, ff=None, topo=None, param=None, prefix='structure', outdir='./bui
...
-    if disulfide is None and not ionize:
-        logger.info('Detecting disulfide bonds.')
-        disulfide = detectDisulfideBonds(mol)
-
-    # Fix structure to match the disulfide patching
-    if not ionize and len(disulfide) != 0:
+    # TODO: Remove this once we deprecate the class
+    from htmd.builder.builder import DisulfideBridge
+    from htmd.molecule.molecule import UniqueResidueID
+    if disulfide is not None and len(disulfide) != 0 and isinstance(disulfide[0], DisulfideBridge):
```
j3mdamas: Maybe additional deprecation warning? (even though it is on the correct place on the `DisulfideBridge` class)

htmd/builder/charmm.py
```diff
@@ -151,21 +152,19 @@ def build(mol, topo=None, param=None, stream=None, prefix='structure', outdir='.
...
     >>> params = ['par/par_all36_prot_mod.prm', './benzamidine.prm', 'par/par_water_ions.prm']
-    >>> disu = [DisulfideBridge('P', 157, 'P', 13), DisulfideBridge('K', 1, 'K', 25)]
+    >>> disu = [['segid P and resid 157', 'segid P and resid 13'], ['segid K and resid 1', 'segid K and resid 25']]
     >>> molbuilt = charmm.build(mol, topo=topos, param=params, outdir='/tmp/build', saltconc=0.15, disulfide=disu)  # doctest: +SKIP
     """
 
     mol = mol.copy()
     _missingSegID(mol)
     _checkMixedSegment(mol)
-    _checkResidueInsertions(mol)
```
j3mdamas: Same comment as in amber.py

htmd/builder/charmm.py
```diff
@@ -251,12 +250,29 @@ def build(mol, topo=None, param=None, stream=None, prefix='structure', outdir='.
         f.write('coordpdb ' + path.join('segments', pdbname) + ' ' + seg + '\n\n')
 
     # Printing out patches for the disulfide bridges
+    # TODO: Remove this once we deprecate the class
+    from htmd.builder.builder import DisulfideBridge
+    from htmd.molecule.molecule import UniqueResidueID
+    if disulfide is not None and len(disulfide) != 0 and isinstance(disulfide[0], DisulfideBridge):
```
j3mdamas: Same comment as amber.py

htmd/builder/amber.py
```diff
@@ -372,10 +388,12 @@ def build(mol, ff=None, topo=None, param=None, prefix='structure', outdir='./bui
...
         for d in disulfide:
+            atoms1 = d[0].selectAtoms(mol, ignore='resname')
+            atoms2 = d[1].selectAtoms(mol, ignore='resname')
             # Convert to stupid amber residue numbering
             uqseqid = sequenceID((mol.resid, mol.insertion, mol.segid)) + mol.resid[0]
-            uqres1 = int(np.unique(uqseqid[(mol.segid == d.segid1) & (mol.resid == d.resid1)]))
-            uqres2 = int(np.unique(uqseqid[(mol.segid == d.segid2) & (mol.resid == d.resid2)]))
+            uqres1 = int(np.unique(uqseqid[atoms1]))
+            uqres2 = int(np.unique(uqseqid[atoms2]))
             f.write('bond mol.{}.SG mol.{}.SG\n'.format(uqres1, uqres2))
```
j3mdamas: Unrelated with this PR, but we're passing the responsibility to tleap to error if the user provides wrong residues (residues that don't even have SG atoms)

htmd/builder/amber.py
```diff
@@ -325,21 +325,37 @@ def build(mol, ff=None, topo=None, param=None, prefix='structure', outdir='./bui
...
+    if not ionize:
+        if disulfide is None:
+            logger.info('Detecting disulfide bonds.')
+            disulfide = detectDisulfideBonds(mol)
+
+        # Fix structure to match the disulfide patching
+        if len(disulfide) != 0:
+            for d in disulfide:
+                # Rename the residues to CYX if there is a disulfide bond
+                atoms1 = d[0].selectAtoms(mol, ignore='resname', indexes=False)
```
j3mdamas: Why the `ignore='resname'`? As long as it is consistent with the actual adding of the disulfide bonds in line 391-2, I guess it's OK, but I don't understand your reasoning. Is it because of the double building? If so, I think it's better to pass all the disulfide detection into the `if not ionize`

htmd/builder/amber.py
```diff
@@ -758,8 +776,9 @@ def run(self, result=None):
         unittest.TestCase.run(self, result) # call superclass run method
 
     def tearDown(self):
-        if self.currentResult.wasSuccessful():
-            shutil.rmtree(self.testDir)
+        pass
```
j3mdamas: Was this intentional or just something you use for debugging and commited by mistake?

htmd/builder/builder.py
```diff
@@ -155,6 +157,15 @@ def embed(mol1, mol2, gap=1.3):
     return mol2
 
 
+def convertDisulfide(mol, disu):
```
j3mdamas: Following my remark above, maybe we want to extend this function to a `validateDisulfide()` function, where we both convert and validate if the passed selections are valid for disulfide bonds? Maybe not erroring (the user may want to do disulfide bonds between a cysteine-like residue which has a weird name), but warn the user is the residue does not have a sulphur and resname does not start with CY*?

htmd/builder/amber.py
stefdoerr: I prefer to only throw one warning instead of two. It is not possible to come to this condition without facing the first warning.

htmd/builder/amber.py
```diff
@@ -255,7 +256,6 @@ def build(mol, ff=None, topo=None, param=None, prefix='structure', outdir='./bui
 
     _missingSegID(mol)
     _checkMixedSegment(mol)
-    _checkResidueInsertions(mol)
```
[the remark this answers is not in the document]
stefdoerr: Nooooo. You are confusing insertion with alternative positions. We need to stick a poster on the wall about this (it confused me a lot too in the past). Insertions are legit residues usually inserted into the normal sequence of the wt protein for the crystalization or whatever other reason.

htmd/builder/amber.py
stefdoerr: Hm yes I could add the check in this PR actually. It's marginal effort and nicer to catch it before shit hits the fan.

htmd/builder/amber.py
stefdoerr: Yes because CYS gets renamed to CYX to create the bridge. Hmmm could we do the disu bonding always post ionize? I doesn't affect the total charge right?

htmd/builder/amber.py
stefdoerr: It doesn't work as I expected it to. It removes the folder even on failure and I couldn't find out the reason so I just removed it for the moment since it complicates debugging a lot if you lose the files.

htmd/builder/builder.py
stefdoerr: Can we do this in a FF generic way? I was planning to do it on the individual builder layer where I know the exact name of the sulfur in a cysteine but if we can assume that you need a single `S*` atom in a `CY*` residue it would also work. But it would not be as exact as you could have some stupid stuff like `SC` instead of `SG` which would crash `tleap`. Maybe better do the check on the individual builders?

htmd/builder/amber.py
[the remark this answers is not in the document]
j3mdamas: I was confusing myself a bit with insertion and altloc, but I believe my comment is still valid: what should we be doing with insertions? In the case of mutations, like 3PTB, we clearly do not want to keep one of the residues for building and should be warning the user, maybe.

htmd/builder/builder.py
```diff
@@ -155,6 +157,15 @@ def embed(mol1, mol2, gap=1.3):
...
+    from htmd.molecule.molecule import UniqueResidueID
+    newdisu = []
+    for d in disu:
+        if not isinstance(d[0], str) or not isinstance(d[1], str):
+            raise RuntimeError('All disulfide selections should be strings')
+        newdisu.append([UniqueResidueID.fromMolecule(mol, d[0]), UniqueResidueID.fromMolecule(mol, d[1])])
```
j3mdamas: I guess if there would be a cystein residue with `altloc`'ed sulphurs, it would immediately fail here (actually, in one of the fail checks inside `UniqueResidueID`) due to returning more than one atom. That's sufficient validation, correct?

htmd/builder/amber.py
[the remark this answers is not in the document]
stefdoerr: No. Insertions are part of the chain and should be there. They just get a insertion number to not ruin the numbering of the residues with respect to their WT sequence. If you remove an inserted residue the crystal structure has a nice gap in space, i.e. you are breaking the protein chain. There is no need for any warnings. Insertions are totally fine residues

htmd/builder/builder.py
stefdoerr: Altlocs are already thrown out by Molecule during reading. There is an option to keep them but then it's up to the user to remove/keep altlocs he wants.

htmd/builder/builder.py
```diff
@@ -171,33 +182,23 @@ def detectDisulfideBonds(mol, thresh=3):
...
-    groups = df.groupby(['resids','segids']).groups
-    for k in groups:
-        if len(df['indexes'][groups[k]].tolist()) != 1:
-            raise RuntimeError('Multiple SG atoms detected in segment {} resid {}. Can\'t guess disulfide bridges.'.format(k[1], k[0]))
+    residues = [UniqueResidueID.fromMolecule(mol, idx=i) for i in idx]
+    for r1 in range(len(residues)):
+        for r2 in range(r1+1, len(residues)):
+            if residues[r1] == residues[r2]:
+                raise RuntimeError('Multiple SG atoms detected in the same residue {}. '
+                                   'Can\'t guess disulfide bridges.'.format(residues[r1]))
```
j3mdamas: OK, maybe I'm thinking too much about robustness (I can't help it), but I think this is possible from what I read here: http://www.bmsc.washington.edu/CrystaLinks/man/pdb/part_62.html: there can be two cysteins with the same residue ID but different insertions, and this will make it error here, while it may not be an error at all.

I think we need to bring the `idx` definition above to work with `UniqueAtomID`. What do you think?

htmd/builder/amber.py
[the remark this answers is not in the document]
j3mdamas: yeah, you're right. See my updated critic below :P

htmd/builder/amber.py
j3mdamas: You're right.

htmd/builder/amber.py
j3mdamas: In the current code (before your changes), it seems like all disulfide machinery is inside the `if not ionize`. It does not affect charge: cystein is either protonated (charge 0) or bound to another cystein (charge 0). Unless the user is working with a deprotonated cystein, but that's an advanced user which should not let disulfides to be defined in an automated way, I guess?

htmd/builder/builder.py
j3mdamas: well, we have the element guessing for sulphur (right?) and I think CY* should work on both FFs (except of the case I already mentioned of cysteine'like residues with weird naming)

htmd/builder/builder.py
j3mdamas: I see. Great!

htmd/builder/charmm.py
```diff
@@ -251,12 +250,29 @@ def build(mol, topo=None, param=None, stream=None, prefix='structure', outdir='.
...
+    if disulfide is not None and len(disulfide) != 0 and isinstance(disulfide[0][0], str):
+        disulfide = convertDisulfide(mol, disulfide)
+
     if disulfide is None:
         disulfide = detectDisulfideBonds(mol)
 
     if len(disulfide) != 0:
         for d in disulfide:
-            f.write('patch DISU {}:{} {}:{}\n'.format(d.segid1, d.resid1, d.segid2, d.resid2))
+            str0 = '{}:{}{}'.format(d[0].segid, d[0].resid, d[0].insertion)
```
j3mdamas: This works in CHARMM? Cool!

htmd/builder/ionize.py
```diff
@@ -157,11 +157,10 @@ def ionizePlace(mol, anion_resname, cation_resname, anion_name, cation_name, nan
     if (nanion + ncation) == 0:
         return newmol
 
-    segname = _getSegname(newmol, segname)
     nions = nanion + ncation
 
-    betabackup = newmol.beta
-    newmol.set('beta', sequenceID((newmol.resid, newmol.segid)))
+    betabackup = newmol.beta.copy()
+    newmol.set('beta', sequenceID((newmol.resid, newmol.insertion, newmol.segid)))
```
j3mdamas: There are other calls to sequenceID around the code. Is this the only relevant one we should be fixing?

j3mdamas:
Oh, just one final remark: I was wondering if the `Unique{Atom,Residue}ID` shouldn't be private classes.

htmd/builder/charmm.py
stefdoerr: Yes! I even tested it and it worked. I was also surprised :P

htmd/builder/ionize.py
stefdoerr: Eh admittedly here it's overkill as it would work even without it. I added it for consistency. But if you find more wrong uses we could review them in another PR

htmd/builder/builder.py
stefdoerr: I don't understand. Two cysteins with insertion code would be different `UniqueResidueID` objects. Or are you talking about `altlocs` (which as I said in the other comment don't exist for us after reading the file)? The comparison here is comparing two `UniqueResidueID` objects, not just their `resid`

j3mdamas: No, actually here I was just messing what the `residues` variable actually meant (and it's equality). Never mind my nonsense!

j3mdamas:
I think these are the three remaining issues:

https://github.com/Acellera/htmd/pull/618#discussion_r180198144
https://github.com/Acellera/htmd/pull/618#discussion_r180207381
https://github.com/Acellera/htmd/pull/618#discussion_r180207818

(just making it easier to track the comments :))

htmd/builder/amber.py
stefdoerr: Ok, fixed now. All disu stuff is post ionize and no need for ignoring the resname

htmd/builder/amber.py
stefdoerr: I reconsidered. I think the builder could add the SG atom if it's missing. This would limit it a bit more. For example if you mutate the residue so it misses the sidechain atoms. I will leave it as it.

htmd/builder/builder.py
stefdoerr: The convert is only there until we deprecate the old way of using the disulfides. Also as I said I'd prefer not to check for the existence of atoms.

stefdoerr:
Ok I responded to the last issues. I added disulfides to only post-ionize and reduced the verbosity of ionize (we already report the added ions in build). For me it's done if @j3mdamas agrees

htmd/builder/builder.py
j3mdamas: I disagree. You still need to convert the user-passed lists to `UniqueResidueID`. Anyway, some refactoring that can be left for later.

j3mdamas:
I'm going to merge.

PR #630, opened 2018-04-04
fix adaptive base for saving PDBs

giadefa:
Simplify the writing of new coordinates that works also when the new coordinate file is a PDB.

The previous optimization is probably not important as this is not a critical part for performance

### DIFF
--- a/htmd/adaptive/adaptive.py
+++ b/htmd/adaptive/adaptive.py
@@ -233,13 +233,8 @@ def _writeInputsFunction(i, f, epoch, inputpath, coorname):
     copytree(currSim.input, newDir, symlinks=False, ignore=ignore_patterns('*.coor', '*.rst', '*.out', *_IGNORE_EXTENSIONS))
 
     # overwrite input file with new one. frameNum + 1 as catdcd does 1 based indexing
-    from htmd.molecule.readers import _MDTRAJ_TRAJECTORY_EXTS
-    import os
-    if os.path.splitext(traj)[1].split('.')[1].lower() in _MDTRAJ_TRAJECTORY_EXTS:
-        # MDtraj trajectory. Unfortunately we need to read the topology to read the trajectory.
-        mol = Molecule(currSim.molfile)
-    else:
-        mol = Molecule()
+
+    mol = Molecule(currSim.molfile)  # Always read the mol file, otherwise it does not work if we need to save a PDB as coorname
     mol.read(traj)
     mol.dropFrames(keep=frameNum)  # Making sure only specific frame to write is kept
     mol.write(path.join(newDir, coorname))


stefdoerr:
Could you possibly just make it check if the `coor` extension is a key in the `_TOPOLOGY_READERS` dictionary of `htmd/molecule/readers.py`? Then if it is, it will read the topology to write it out again, otherwise not. I remember that this change was to support use cases for MD engines where we don't have topology readers but still can make it work if they write in a compatible trajectory format, i.e. `pmemd`. Also reduces warning-spam from the readers and makes it faster for stuff like any sort of fake or synthetic adaptive where you might spawn hundreds of simulations.

giadefa:
The way it is, it always read the mol file, so it would work in all cases

giadefa:
if htmd cannot read the mol file, then it cannot do anything really even if you can write out a coordinate file. The solution would be to add the missing reader

giadefa:
second motivation is that if we do checks like that, we actually make the code unusable because to fix a corner case another one is broken (as in this case). Code should be minimalistic.

giadefa:
@j3mdamas  The assignee is the one that should approve it or the one that made the PR?

stefdoerr:
We agreed to put as assignee the one that writes the code and the reviewer is the reviewer

On April 4, 2018 1:52:30 PM GMT+03:00, Gianni De Fabritiis <notifications@github.com> wrote:
>@j3mdamas  The assignee is the one that should approve it or the one
>that made the PR?
>
>-- 
>You are receiving this because your review was requested.
>Reply to this email directly or view it on GitHub:
>https://github.com/Acellera/htmd/pull/630#issuecomment-378559936

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

stefdoerr:
Ok fine. We don't use the synthetic adaptive anymore anyway and this would make it more flexible.

On April 4, 2018 2:34:21 PM GMT+03:00, stefdoerr@gmail.com wrote:
>We agreed to put as assignee the one that writes the code and the
>reviewer is the reviewer
>
>On April 4, 2018 1:52:30 PM GMT+03:00, Gianni De Fabritiis
><notifications@github.com> wrote:
>>@j3mdamas  The assignee is the one that should approve it or the one
>>that made the PR?
>>
>>-- 
>>You are receiving this because your review was requested.
>>Reply to this email directly or view it on GitHub:
>>https://github.com/Acellera/htmd/pull/630#issuecomment-378559936
>
>-- 
>Sent from my Android device with K-9 Mail. Please excuse my brevity.

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

stefdoerr:
We need to merge the master in here to pass the tests still

j3mdamas:
master needs to be merged into this in order to execute the ci tests

j3mdamas:
@giadefa you can merge now, the master has now the fixes for passing travis

j3mdamas:
I meant merge master into this PR.

j3mdamas:
I'll fix it, don't worry.

PR #636, opened 2018-04-04
Fix htmdlib

### LINKED ISSUE #623: htmd lib for xtc writing is not updated
from htmd.ui import *
mol = Molecule('3ptb')
mol.coords = np.zeros((mol.numAtoms, 3, 100), dtype=np.float32)
mol.write('test.xtc')
mol2 = Molecule('3ptb')
mol2.read('test.xtc')
mol2.coords.shape

this simple code shows it

j3mdamas:
Will close #623


htmd/tests/run_inline_file_tests.py
```diff
@@ -70,16 +70,23 @@ def containsTests(fname):
...
+        process = Popen('export PYTHONHASHSEED=1; {} {}'.format(pythonexe, f), shell=True, stdout=PIPE, stderr=STDOUT,
+                        universal_newlines=True)
     else:
-        process = run('{} {}'.format(pythonexe, f), shell=True)
+        process = Popen('{} {}'.format(pythonexe, f), shell=True, stdout=PIPE, stderr=STDOUT,
+                        universal_newlines=True)
+    while True:
+        out = process.stdout.readline()
+        print(out.strip())
+        if len(out) == 0:
```
stefdoerr: If you are sure this is only 0 when finished ok. What would happen if it just didn't print anything between two iterations?

htmd/tests/run_inline_file_tests.py
```diff
@@ -70,16 +70,23 @@ def containsTests(fname):
...
-        process = run('export PYTHONHASHSEED=1; {} {}'.format(pythonexe, f), shell=True)
+        process = Popen('export PYTHONHASHSEED=1; {} {}'.format(pythonexe, f), shell=True, stdout=PIPE, stderr=STDOUT,
+                        universal_newlines=True)
     else:
-        process = run('{} {}'.format(pythonexe, f), shell=True)
+        process = Popen('{} {}'.format(pythonexe, f), shell=True, stdout=PIPE, stderr=STDOUT,
+                        universal_newlines=True)
+    while True:
+        out = process.stdout.readline()
+        print(out.strip())
```
stefdoerr: Why strip?

htmd/tests/run_inline_file_tests.py
j3mdamas: there are some tests that don't print anything, and it breaks the loop in the correct manner. I've ran locally and on Travis, and it prints nicely. But I admit maybe I should have used https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate here.
Since this is a major bug, and this works for now, I'll merge and I'll open an issue on improving this printing.

htmd/tests/run_inline_file_tests.py
j3mdamas: It comes out with `\n` in the end.

PR #635, opened 2018-04-04
redid the PR of MetricCoordinate

stefdoerr:
there was some sort of conflict

PR #637, opened 2018-04-04
abstract classes need ABCMeta

stefdoerr:
This fixes the lack of enforcing of the abstract methods of `Projection` which lead to `MetricCoords` passing the tests (but would have crashed if used with `Metric` since it lacked the `_precalculate` method)


j3mdamas:
I will complete this PR, as I think we need to fix it across HTMD. I.e.:

* AdaptiveBase should also have it
* QMBase is using it differently

j3mdamas:
Also, this PR depends on this one: https://github.com/Acellera/htmd/pull/635 (as soon as this gets merged, master will break in CI)

j3mdamas:
Merged new master with #635.

About my comment https://github.com/Acellera/htmd/pull/637#issuecomment-378766260, I'll do:

* Update `AdaptiveBase` to have `metaclass=ABCMeta`
* I'll leave `QMBase` as it is. Apparently, from python 3.4, they implemented the ABC helper class (https://docs.python.org/3/library/abc.html#abc.ABC), and from what I read, it could be better than using `metaclass=ABCMeta`. Let's post-pone that uniformization.

PR #651, opened 2018-04-15
Fixes old "from htmd import *" docs

### LINKED ISSUE #622: from htmd import *
Still present in docs
https://software.acellera.com/docs/latest/htmd/userguide/adaptive-sampling-explained.html

do a search

stefdoerr:
Closes #622

### DIFF
--- a/doc/source/developers/guidelines.rst
+++ b/doc/source/developers/guidelines.rst
@@ -27,7 +27,7 @@ HTMD code:
 .. code:: python
 
     import htmd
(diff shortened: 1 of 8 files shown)


stefdoerr:
I doubt this needs much of a review but anyway have a look at it. There are some more uses in the old docs and tutorials but I thought it pointless to fix those as they should be removed. Really, why do we keep the old docs and tutorials around? It's not like git doesn't keep them in history.

PR #663, opened 2018-04-20
Added 'material' and 'selupdate' in Representation 

cuzzo87:
added representation options:
ONLY VMD
* material --> for chosing material style
* selupdate --> for enabling selection update at each frames

### DIFF
--- a/htmd/molecule/molecule.py
+++ b/htmd/molecule/molecule.py
@@ -1835,7 +1835,7 @@ class Representations:
     >>> mol = tryp.copy()
     >>> mol.reps.add('protein', 'NewCartoon')
     >>> print(mol.reps)                     # doctest: +NORMALIZE_WHITESPACE
-    rep 0: sel='protein', style='NewCartoon', color='Name'
+    rep 0: sel='protein', style='NewCartoon', color='Name', material='Opaque'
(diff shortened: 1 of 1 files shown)


stefdoerr:
Well it's a very internal thing since he didn't modify the Molecule.view method. So I guess it's fine if he really needs it. We will eventually see how we deal with this when we switch to NGL but I assume NGL should also support materials at some point if it doesn't already.

j3mdamas:
@stefdoerr the `Molecule.view()` was not changed, but the `Representation` class yes, and that's accessible to the user through the `add` method, which was also changed.

Actually, the `selupdate` and `material` should be two separate issues. @cuzzo87, this is for the video stuff you're doing, right? Can't this PR be killed and you add these to that video stuff, and we decide then?

stefdoerr:
Yeah but no one uses the add and representation class.

j3mdamas:
"no one". It's a public method. The other day I saw a ipynb from Shozeb and he was using them.

stefdoerr:
Hah okay :D

cuzzo87:
For me is okay. The only "problem" is that also the video maker PR is working only for VMD. NGLview does not allow to use the required NGL method. I am waiting for them a possible implementation.

PR #665, opened 2018-04-21
Silence pyemma version warnings

stefdoerr:
Really the other PR was overkill. Those warnings are only thrown on the first import. So we might as well just import it once with a wrapper in `htmd.ui` and be done with it.

### DIFF
--- a/htmd/ui.py
+++ b/htmd/ui.py
@@ -76,7 +76,7 @@ import warnings
 # Get rid of pyemma version warnings
 with warnings.catch_warnings():
     warnings.simplefilter('ignore', category=UserWarning)
-    from pyemma import coordinates
+    import pyemma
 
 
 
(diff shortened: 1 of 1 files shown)


stefdoerr:
@j3mdamas this is another very trivial PR for a 1 second review. The only question is if we care about the flows. If someone import directly Model or TICA without using `ui` it will still throw the pyemma warning. But I personally don't think we care that much about this case

stefdoerr:
I mean I could silence the warning in those files as well but I don't feel like splattering this stuff all around the HTMD code

j3mdamas:
I assume we do not want to do it on the `__init__.py`, right?
I think for the Flows, we could put such a code there as well.

stefdoerr:
No, it would slow down all import in the init. Thanks!

PR #674, opened 2018-04-30
slightly easier use of plotTimescales

stefdoerr:


stefdoerr:
Yes those heuristics are probably crap. I just didn't want to modify current behaviour to not get into a huge discussion.

j3mdamas:
agreed. accepted and merged :)

PR #675, opened 2018-05-01
bug in plotting function

stefdoerr:
@j3mdamas you can pretty much auto-accept this one. There is no `.data` in `MetricData`. It was a mistake in copy pasting some code from `Model`

### DIFF
--- a/htmd/model.py
+++ b/htmd/model.py
@@ -924,7 +924,7 @@ def _loadMols(self, rel, molfile, wrapsel, alignsel, refmol, simlist):
     return mol
 
 
-def getStateStatistic(model, data, states, statetype='macro', weighted=False, method=np.mean, axis=0, existing=False, target=None):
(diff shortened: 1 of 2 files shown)


stefdoerr:
Well okay it ended up growing a bit more :D Still it's some necessary bug fixes with a few improvements included

htmd/model.py
```diff
@@ -849,7 +849,11 @@ def plotFES(self, dimX, dimY, temperature, states=False, s=10, cmap=None, plot=T
         s : float
             Marker size for states.
         cmap :
-            Matplotlib colormap.
+            Control both countscmap and statecmap
```
j3mdamas: I guess it just needs a better doc? `Sets the Matplotlib colormap for both `fescmap` and `statescmap``?

stefdoerr: ok

stefdoerr: Writing an example as well

PR #676, opened 2018-05-01
aligned bars correctly to kinetics plots

stefdoerr:
The alignment was off and it was confusing me


stefdoerr:
This must have been a fix by matplotlib in the positioning of bar plots that made my old fix wrong again.

PR #678, opened 2018-05-02
Smallmol unitest and bug fixes

cuzzo87:
* Implemented unitest for the smallmol package.
 * added calculateAngle mostly for tests
 * fix a bug in determine coordinates for element to be attached to an atom with hybirdization SP


j3mdamas:
OK, so this is the next target to merge. @cuzzo87 and I are going to do it on Wednesday.

PR #541, opened 2018-01-31
Parameterize: new DFT functional and basis sets 

raimis:
**Do not merge before we have updated to Psi4 1.2** (https://github.com/psi4/psi4/milestone/3)

- [x] 6-311G** and 6-311++G** basis
- [x] wB97X-D functional
  - [x] Tests

PR #679, opened 2018-05-03
Parameterize: custom calculator

raimis:
**Do not merge!**

Depend on #541


j3mdamas:
I need to protect the qmml module import :D

j3mdamas:
@raimis I shut down the custom calculator doctest, as the qmml module is not made available on Travis to test. The doctest is also outdated (does not take the parameterization re-factoring we did)

Issue #681, opened 2018-05-11
Add mmCIF reading capabilities


j3mdamas:
Large structures aren't available as PDB on RCSB:

http://www.rcsb.org/structure/5Y6P

stefdoerr:
I'm on it, don't worry

PR #682, opened 2018-05-12
Added mmCIF reader

stefdoerr:
I took the free code from PDB for parsing mmCIF http://mmcif.wwpdb.org/docs/sw-examples/python/html/ and added my own function for converting to a Molecule. Seems to work fine on one case that I tested `1ffk.cif`.
It probably needs some more work to read more stuff like box size, multiple frames or whatever else is supported. But for the moment it's fun to look at huge RNA structures :laughing: 

Will close https://github.com/Acellera/htmd/issues/681


stefdoerr:
Finished the reader based on the official correspondence of fields:
http://mmcif.wwpdb.org/docs/pdb_to_pdbx_correspondences.html

stefdoerr:
Still won't read multiple models but I am not sure it's that necessary anyway. We pretty much never read multiple models.

stefdoerr:
Also I won't bother with the writer for now. It's doable but I don't see anyone wanting to use a mmCIF writer in the close future.

PR #687, opened 2018-05-17
added reporting to the wait method of queues to show progress of calc…

stefdoerr:
…ulations


stefdoerr:
No wait I am rethinking it a bit

stefdoerr:
Okay I think now the logic makes more sense. If the reporting time is shorter than the sleeping time then the sleeping time should be equal to the reporting time. If the reporting time is longer than the sleeping time it should report every round(reporttime/sleeptime) iterations.

stefdoerr:
Thought I might as well add a callback argument for playmolecule in case Gerard wants to make it look nice

j3mdamas:
all of these new need the preparation fix which is not merged into master yet

stefdoerr:
I know

stefdoerr:
@j3mdamas can ye review pliz?

stefdoerr:
Although this one actually would have needed some testing :P

j3mdamas:
do you want me to test it?

stefdoerr:
If you feel like it :P You would need some fake sleep-WU to test it

PR #701, opened 2018-05-30
Added support for multimol Mol2 files

stefdoerr:
Updated the old mol2 reader. Made it able to read multimol files.
Molecule class will still not be able to read multimol files since this will require a decent redesign of stuff. But at least if someone wishes to use the reader directly to read multiple mol2 molecules he can. 
I'll probably make another PR later that allows `Molecule.read` to read and return multiple molecules.

### DIFF
--- a/htmd/molecule/molecule.py
+++ b/htmd/molecule/molecule.py
@@ -806,7 +806,7 @@ class Molecule:
         self.moveBy(-com)
         self.moveBy(loc)
 
-    def read(self, filename, type=None, skip=None, frames=None, append=False, overwrite='all', keepaltloc='A', guess=None, guessNE=None, _logger=True):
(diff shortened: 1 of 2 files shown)


j3mdamas:
@stefdoerr can you setup a priority for this? thanks!

stefdoerr:
Considering it's a small change, the sooner the better because you know how much fun it is to merge branches and updates :P Let's see though if the tests pass. But really it's just a 5 minutes review max.

stefdoerr:
It's not urgent though in any sense because I doubt it will manage to make
it to the current smallmol without some hacking

stefdoerr:
Took you too long ;) Now you will suffer the consequences. I close this to open soon a much bigger PR which refactors the readers and Molecule.read.

PR #703, opened 2018-05-31
Smallmol unitest and bug fixes (part 2)

j3mdamas:
This deprecates #678.

It has the same implementation, but with the new htmd-data.


j3mdamas:
As soon as it passes Travis, I'll merge it. I've ran the unittests locally, and they passed. I'm not going to do a big review, because it makes  the remaining SmallMol PRs from @cuzzo87 (pre-new-htmd-data) much harder to solve. Since SmallMol is beta, it's OK. In the end, I do a mass review over it and do a PR on it.

j3mdamas:
Just time out. Push and half of PR passed, it's OK. Merging.

PR #704, opened 2018-05-31
SmallMol: Retrieve capabilities

j3mdamas:
This is a remake of https://github.com/cuzzo87/htmd/tree/smallmol_retrieveLigand to include the changes in htmd-data.

@cuzzo87 this one you really need to review carefully. You did not change anything in the data for this PR, right?

This is going to be merged fast, as soon as it passes tests.

Things missing:
- [] @cuzzo87 needs to confirm this is functional
- [] @j3mdamas needs to add `chembl_webresource_client` to list of dependencies

PR #705, opened 2018-06-01
SmallMol: Retrieve capabilities (version 2)

j3mdamas:
Remaking https://github.com/Acellera/htmd/pull/704

Missing:

- [x] @cuzzo87 needs to make `chembl_webresource_client` work like biopython in Molecule


j3mdamas:
Great, all is good here. @cuzzo87, could you do a final review over it?

PR #711, opened 2018-06-11
Refactoring of Molecule file reading

stefdoerr:
As promised here comes the big PR.
So as per the discussion we had on multi-mol MOL2 file reader. 
We needed the readers to be able to return multiple molecules. In readers.py I have the Topology and Trajectory classes which I only use temporarily to collect all the data in the file because appending to lists is faster than appending to numpy arrays which is what Molecule uses. 

So then I needed a way to convert a Topology+Trajectory to a Molecule. Until now this was done inside the Molecule class. This however creates the problem that if you read a file with multiple molecules you need to instantiate multiple Molecule objects which could not be done in `Molecule.read`. So I moved the Molecule creation part out of Molecule now to a MolFactory class which creates Molecule objects. Then the readers can use that to return one or more Molecules. This allows us now to read multiple molecules from a single file by using directly the file reader.

Hence the `Molecule.read` now gets from the readers one or more `Molecule` objects which it then tries to merge into itself.

And to answer the probably inevitable question, the `MolFactory` is not meant to be instantiated. It's a static class of the "factory/builder" pattern https://en.wikipedia.org/wiki/Builder_pattern. It probably doesn't follow the patterns to the T but I hope you don't mind. Yes, I could have made it a function instead of a class but I thought it nicer to group the functions together in that way. Your taste may vary.

If this takes too long to merge it will probably become a mess to merge later since it changes some core stuff in Molecule so please @j3mdamas review the soonest possible.

### DIFF
--- a/htmd/molecule/molecule.py
+++ b/htmd/molecule/molecule.py
@@ -881,6 +881,7 @@ class Molecule:
             if ext not in _ALL_READERS:
                 raise ValueError('Unknown file type with extension "{}".'.format(ext))
             readers = _ALL_READERS[ext]
+            mol = None
             for rr in readers:
                 try:
                     mol = rr(fname, frame=frame, topoloc=tmppdb, **kwargs)
@@ -888,6 +889,12 @@ class Molecule:
                     continue
                 else:
                     break
+
+            if mol is None:
+                raise RuntimeError('No molecule read from file {} with any of the readers {}'.format(fname, readers))
+
+            if isinstance(mol, list):
+                raise AssertionError('Reader {} should not return multiple molecules. Report this error on github.')
 
             if self.numAtoms != 0 and mol.numAtoms != self.numAtoms:
                 raise ValueError('Number of atoms in file ({}) mismatch with number of atoms in the molecule '
--- a/htmd/molecule/molecule.py
+++ b/htmd/molecule/molecule.py
@@ -882,36 +882,29 @@ class Molecule:
             readers = _ALL_READERS[ext]
             for rr in readers:
                 try:
-                    to, tr = rr(fname, frame=frame, topoloc=tmppdb, **kwargs)
+                    newmol = rr(fname, frame=frame, topoloc=tmppdb, **kwargs)
                 except FormatError:
                     continue
                 else:
                     break
 
-            if tr is not None:
-                self._keepFrame(tr, frame)
-                self._checkCoords(tr, rr, fname)
-                # TODO: Get rid of this if by moving it to a function
-                if ext in _TRAJECTORY_READERS and frame is None:
-                    # Writing hidden index file containing number of frames in trajectory file
-                    if os.path.isfile(fname):
-                        self._writeNumFrames(fname, tr.coords[0].shape[2])
-                    ff = range(np.size(tr.coords[0], 2))
-                    #tr.step = tr.step + traj[-1].step[-1] + 1
-                elif frame is None:
-                    ff = [0]
-                elif frame is not None:
-                    ff = [frame]
-                else:
-                    raise AssertionError('Should not reach here')
-                tr.fileloc = [[fname, j] for j in ff]
+            if newmol.numFrames != 0:
+                if frame is not None and newmol.numFrames > 1:
+                    newmol.dropFrames(keep=frame)
+
+                if self.numAtoms != 0 and newmol.numAtoms != self.numAtoms:
+                    raise ValueError(
+                        'Number of atoms in trajectory ({}) mismatch with number of atoms in the molecule ({})'.format(
+                            newmol.numAtoms, self.numAtoms))
+
+                # TODO: Append trajectories
                 traj += tr
 
-            if to is not None:
-                self._parseTopology(to, fname, overwrite=overwrite, _logger=_logger)
+            if newmol.numAtoms != 0:  # TODO: THIS WILL NOT WORK. THIS CHECKS ALSO THE TRAJ FOR NUMATOMS!!!!
+                self._parseTopology(to, fname, overwrite=overwrite, _logger=_logger) # TODO: Merge topologies!
 
         if len(traj.coords) != 0:
-            self._parseTraj(traj, skip=skip)
+            self._parseTraj(traj, skip=skip)P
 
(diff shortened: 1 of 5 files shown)


j3mdamas:
On Wednesday. Sooner only if I get access to a computer today or tomorrow.
João

j3mdamas:
Looking at it now

j3mdamas:
General comment: I wonder if the strategy taken here should be applied in the `SmallMolLib`

stefdoerr:
In which sense? I think currently SmallMol reads using mostly rdkit. But
yes smallmollib could now use the readers directly and convert the Molecule
to an rdkit mol

htmd/molecule/molecule.py
```diff
@@ -310,14 +313,17 @@ def insertappend(index, data1, data2, append):
...
-            for k in self._atom_fields:
+            for k in self._atom_and_coord_fields:
                 if k == 'serial':
                     continue
                 data2 = mol.__dict__[k]
                 if mol.__dict__[k] is None or np.size(mol.__dict__[k]) == 0:
                     data2 = self._empty(mol.numAtoms, k)
                 self.__dict__[k] = insertappend(index, self.__dict__[k], data2, append)
             self.serial = np.arange(1, self.numAtoms + 1)
+            # Reset the box to zeros as you cannot keep box size after inserting atoms
```
j3mdamas: Why not? The user could have setup a particular box size and be using some protocol to insert molecules inside it (imagine some solvate procedure with organic molecules). Or am I missing the point here?

htmd/molecule/molecule.py
```diff
@@ -806,7 +812,7 @@ def center(self, loc=(0, 0, 0), sel='all'):
         self.moveBy(-com)
         self.moveBy(loc)
 
-    def read(self, filename, type=None, skip=None, frames=None, append=False, overwrite='all', keepaltloc='A', guess=None, guessNE=None, _logger=True):
+    def read(self, filename, type=None, skip=None, frames=None, append=False, overwrite='all', keepaltloc='A', guess=None, guessNE=None, _logger=True, **kwargs):
```
j3mdamas: but I thought you hated kwargs :p

htmd/molecule/molecule.py
```diff
@@ -1279,6 +1232,15 @@ def write(self, filename, sel=None, type=None, **kwargs):
             raise IOError('Molecule cannot write files with "{}" extension yet. If you need such support please notify '
                           'us on the github htmd issue tracker.'.format(ext))
 
+    def _emptyTopo(self, numAtoms):
+        for field in Molecule._atom_fields:
```
j3mdamas: shouldn't it be `Molecule._topo_fields`?

htmd/molecule/molecule.py
```diff
@@ -1279,6 +1232,15 @@ def write(self, filename, sel=None, type=None, **kwargs):
...
+            self.__dict__[field] = self._empty(numAtoms, field)
+
+    def _emptyTraj(self, numAtoms):
```
j3mdamas: shouldn't it initialize all `Molecule._traj_fields`?

htmd/molecule/molecule.py
stefdoerr: No, we don't initialize the ['bonds', 'bondtype', 'angles', 'dihedrals', 'impropers', 'crystalinfo'] fields. Just the ones with atom information. The others can be empty.

htmd/molecule/molecule.py
stefdoerr: We cannot initialize `'fileloc', 'step', 'time'` fields

stefdoerr: because they are not np.arrays

htmd/molecule/molecule.py
```diff
@@ -882,36 +883,23 @@ def read(self, filename, type=None, skip=None, frames=None, append=False, overwr
...
-                    #tr.step = tr.step + traj[-1].step[-1] + 1
-                elif frame is None:
-                    ff = [0]
-                elif frame is not None:
-                    ff = [frame]
-                else:
-                    raise AssertionError('Should not reach here')
-                tr.fileloc = [[fname, j] for j in ff]
-                traj += tr
+            if self.numAtoms != 0 and mol.numAtoms != self.numAtoms:
```
j3mdamas: I wonder what the `self.numAtoms != 0` is adding to the if-clause here, but I guess it's ok

htmd/molecule/molecule.py
```diff
@@ -882,36 +883,23 @@ def read(self, filename, type=None, skip=None, frames=None, append=False, overwr
...
+                                 '({})'.format(mol.numAtoms, self.numAtoms))
 
-            if to is not None:
-                self._parseTopology(to, fname, overwrite=overwrite, _logger=_logger)
+            if frame is not None:
+                mol.dropFrames(keep=frame)
 
-        if len(traj.coords) != 0:
-            self._parseTraj(traj, skip=skip)
+            newmols.append(mol)
```
j3mdamas: mol can be a `list()`, so I suggest changing this to:
```python
if isinstance(mol, list):
    newmols.extend(mol)
else:
    newmols.append(mol)
```

htmd/molecule/molecule.py
stefdoerr: This covers the case that the original Molecule is empty. Like doing `mol = Molecule(); mol.read(x)`. Then we want to not throw an error.

htmd/molecule/molecule.py
stefdoerr: Good point but as we discussed Molecule should read single molecules from files. I will remove the kwargs and force the mol2 option for single molecule

stefdoerr: Actually I'm leaving the kwargs in case some readers in the future require it. It's a bit undocumented and hacky I guess but gives more flexibility if we want to solve some issue of a user for example and doesn't really hurt at the moment. We already have one in the .write.
mol2read by default reads only the first molecule so I think it's ok. I will add a check though that the reader didn't return multiple molecules.

stefdoerr:
Yes you have a point. We assume that multi-model PDB files are trajectories with the same topology while in the MOL2reader we assume that they are different molecules.

Issue #712, opened 2018-06-11
VMD-style select doesn't work when resname is all-digits


lyu18:
Hello,

I found the VMD-style select won't work when seleting residues whose `resname` are all digits (i.e. no Roman character.)

Example:

PDB entry `3udh` has a small molecule ligand whose `resname` is `091`. Say I want to see remove everything except this ligand,

```python
from htmd.molecule.molecule import Molecule
mol = Molecule('3udh')
mol.filter(sel='resname 091')
```

This returns:

```
2018-06-11 14:02:19,552 - htmd.molecule.molecule - INFO - Removed 3464 atoms. 0 atoms remaining in the molecule.
```

*Everything is removed,* i.e. nothing is selected.

However, I know the `resid` for this ligand is `501`, and it works well when selecting it with `resid`:

```python
mol = Molecule('3udh')
mol.filter(sel='resid 501')
```

```
2018-06-11 14:08:40,086 - htmd.molecule.molecule - INFO - Removed 3450 atoms. 14 atoms remaining in the molecule.
```

it also works well for ligands whose `resname` are not all-digits. For example, for another ligand in this PDB entry whose name is `EDO`:

```python
mol = Molecule('3udh')
mol.filter(sel='resname EDO')
```

```
2018-06-11 14:10:03,715 - htmd.molecule.molecule - INFO - Removed 3440 atoms. 24 atoms remaining in the molecule.
```

stefdoerr:
Hmmmm that's curious. Could you tell me what `mol.resname.dtype` gives?

lyu18:
Here it is:

```python
from htmd.molecule.molecule import Molecule
mol = Molecule('3udh')
print(mol.resname.dtype)
```

```
object
```

stefdoerr:
Yes I just checked. Seems like the issue is with numerical resnames starting with 0. The pandas reader cuts off the  zeros because it tries to cast it to integer. I'll try to see what we can do as this coincides with another issue with our PDB reader so I need to see if it's better to fix it or ditch pandas and write my own parser.

lyu18:
That makes sense. Thanks Stefan.

stefdoerr:
I see that pandas finally allows dtypes for read_fwf. So it should be trivial to do now! I'll give it a try. I'll keep this issue open till it's fixed if you don't mind :)

stefdoerr:
[...]

PR #713, opened 2018-06-14
Improvements for PDB reader

stefdoerr:
This PR should fix the following two issues:
https://github.com/Acellera/htmd/issues/708 Hydrogen detected as Hahnium
https://github.com/Acellera/htmd/issues/712 Reading resnames starting with zeros


stefdoerr:
This one is actually easier @j3mdamas :P Won't hurt as much as the last one.

stefdoerr:
Depends on https://github.com/Acellera/htmd-data/pull/4 being merged as now our PDBreader will fail on broken PDB files (which it should)

j3mdamas:
The best (and worst) thing about the submodule, is that we can merge this and it will still reference the right commit (even though you did well with the branch and PR there, that's what I've been doing as well)

j3mdamas:
merging it

PR #714, opened 2018-06-14
Fixes one bug in ringpenetration and many errors in lgtm

stefdoerr:
Trying to reduce errors in LGTM. Most of them were either from non-object classes or from cyclic imports caused by `from htmd.ui import *` in some `__main__` sections.
Might need another PR later but I wanted to see how many errors this gets rid of :)

### DIFF
--- a/htmd/membranebuilder/ringpenetration.py
+++ b/htmd/membranebuilder/ringpenetration.py
@@ -96,10 +96,10 @@ def _checkAtomsPenetrating(coords2, bonds2, ringcoords, axis, ringcom, maxd, box
                 continue
 
             d = 0
-            for i in range(0, ringcoords.shape[0]):
-                p1 = ringcoords[i] - p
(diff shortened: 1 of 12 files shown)


htmd/membranebuilder/ringpenetration.py
```diff
@@ -96,10 +96,10 @@ def _checkAtomsPenetrating(coords2, bonds2, ringcoords, axis, ringcom, maxd, box
                 continue
 
             d = 0
-            for i in range(0, ringcoords.shape[0]):
-                p1 = ringcoords[i] - p
+            for k in range(0, ringcoords.shape[0]):
```
j3mdamas: ouch, this one is a classic :p

stefdoerr: yeah, I don't want to think how it worked ;P

PR #715, opened 2018-06-15
Parmed 3 fix

stefdoerr:
This makes HTMD compatible with parmed 3+ but breaks compatibility with parmed 2.7.
It will not pass the tests here since we would need to update parmed on acellera channel but I tested it locally and it passes all tests. 

https://github.com/ParmEd/ParmEd/issues/990
As per this issue I had to implement improper center detection to be compatible with the new way it stores impropers.


stefdoerr:
I made it work for both parmed 2.7 and 3 now! So we can merge with no issues. Also simplified a bit the code and the usage

j3mdamas:
This now depends on #729, no?

stefdoerr:
Nah, just on the docstrings which are not evaluated on testing.

j3mdamas:
it's merged in any case :)

PR #718, opened 2018-06-21
Fix AceCloudQueue

j3mdamas:
A single commit with 2 fixes and 1 change.

This was tested to work with AdaptiveMD, and it works great.

@stefdoerr can you review it fast so I can release today or tomorrow already with these changes. Thanks!

### DIFF
--- a/htmd/queues/acecloudqueue.py
+++ b/htmd/queues/acecloudqueue.py
@@ -48,7 +48,8 @@ class AceCloudQueue(SimQueue, ProtocolInterface):
         self._arg('groupname', 'str', 'The name of the group of simulations you want to submit. If none is given, '
                                       'a randomly generated string will be used instead.', None, val.String())
         self._arg('datadir', 'str', 'The directory in which to retrieve your results.', None, val.String())
-        self._arg('instancetype', 'str', 'Instance type', 'g2.2xlarge', val.String(), valid_values=('g2.2xlarge', 'r4.large', 'p2.xlarge'))
+        self._arg('instancetype', 'str', 'Instance type', 'p2.xlarge', val.String(),
+                  valid_values=('g2.2xlarge', 'r4.large', 'p2.xlarge'))
         self._arg('hashnames', 'bool', 'If True, each job will have a name created from the hash of its directory '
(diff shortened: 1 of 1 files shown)


htmd/queues/acecloudqueue.py
```diff
@@ -91,7 +92,7 @@ def submit(self, dirs):
                 import hashlib
                 name = hashlib.sha256(os.path.abspath(d).encode('utf-8')).hexdigest()[:10]
 
-            runscript = self._getRunScript(d)
+            runscript = self._hackRunScript(d)
```
stefdoerr: :laughing:

htmd/queues/acecloudqueue.py
```diff
@@ -169,6 +167,18 @@ def _createJobScript(self, fname, workdir, runsh):
             f.write('bash run.sh')
         os.chmod(fname, 0o700)
 
+    def _hackRunScript(self, directory):
+        import fileinput
+
+        bak_suffix = '.acecloud.bak'
```
stefdoerr: > If the backup parameter is given (typically as backup='.<some extension>'), it specifies the extension for the backup file, and the backup file remains around; by default, the extension is '.bak' and it is deleted when the output file is closed.

Are you sure it recognizes that `.acecloud.bak` is not `.bak` and deletes it? I assume you checked but just in case I'm asking

htmd/queues/acecloudqueue.py
j3mdamas: I know it's ugly, but it's easier than to "fix" the AceCloud image

htmd/queues/acecloudqueue.py
j3mdamas: well, it does leave it behind. even when I was using just `.bak` it was leaving it behind. Do you think it's a bug and I should do something safer? I know there are several ways of accomplishing what I am doing here but this `fileinput` stuff seemed to be the easier (and the closest to a `sed`)

stefdoerr: Nah I guess it differentiates between default and user-passed value somehow? If it stays it's ok

Issue #719, opened 2018-06-22
Charmm Build create structure with missing atoms


cuzzo87:
Using a ligand rtf and parameters provided by parameterize to build a structure, some atoms were not placed. The log error  (example):
```
....
psfgen) ERROR: Missing atoms for bond C17(0) O3(0) in residue
...
psfgen) Warning: failed to set coordinate for atom O3
...
```

The reason is a missing space between the atom type and the charge in the rtf file for the atom O3 (example):
```
...
ATOM   O3 O311x0-0.124400
...
```
By adding the space to all these atoms in the rtf the structure was perfectly built

stefdoerr:
https://github.com/Acellera/htmd/pull/722
This should fix it. You would need to rerun the parameterization on the ligand though to rewrite the RTF file.

PR #722, opened 2018-06-22
Fix for parameterize RTF writer

stefdoerr:
fixes problems with atom types overflowing in the rtf writer causing pr…oblems in psfgen

https://github.com/Acellera/htmd/issues/719


stefdoerr:
The CHARMM RTF specification does not define any column format so it's space separated but for prettiness and backwards compatibility I made the column format more flexible

j3mdamas:
When it passes, we can merge.

PR #736, opened 2018-07-06
Some minor lgtm fixes plus telling it that the project is python 3

stefdoerr:
https://help.semmle.com/lgtm-enterprise/user/help/analysis-faqs.html#why-are-there-false-alerts-in-my-python-project

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,15 +4,18 @@ sudo: required
 dist: trusty
 group: deprecated-2017Q4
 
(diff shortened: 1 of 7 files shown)


j3mdamas:
Apparently, the two parameterize tests that I was having problems with are also failing here, so it seems unrelated with joblib version pinning. Let me merge that branch into master (when it passes), and then master can be merged into this branch.

PR #739, opened 2018-07-12
fixes align jit (should close #738)

### LINKED ISSUE #738: maximalSubstructureAlignment does not work
I was trying this function and I got an argument type error. The _pp_align function in Molecule class expects the selection of both molecules (reference and query) to be boolean arrays. This is not the case for the output of maximalSubstructureAlignment, which returns two list of indexes. If one changes the signature of the __pp_align function to: 

```
@jit('float32[:, :, :](float32[:, :, :], float32[:, :, :], int64[:], int64[:], int64[:], int64)', nopython=True,
     nogil=True)
```

and converts the two list of indexes from maximalSubstructureAlignment into two numpy arrays, it works. Converting these indexes to boolean arrays does not solve it, as the pairing between the nodes in reference and query breaks.

Script to reproduce the result:

```
from htmd.ui import *
from htmd.molecule.graphalignment import maximalSubstructureAlignment

reference_crystal = Molecule('2Y02')
reference_crystal.filter('protein or resname WHJ')
reference_ligand = reference_crystal.copy()
reference_ligand.filter('resname WHJ')
#there are two ligands with the same resname, so we select one of them
reference_ligand.filter('residue 1') 

#extract other ligand from this congeneric series from protein 2Y03
ligand_to_align = Molecule('2Y03')
ligand_to_align.filter('resname 5FW')
#residue 1 happens to be already align with the reference ligand, so we use residue 0
ligand_to_align.filter('residue 0') 

reference_crystal.view('alrefprot')
ligand_to_align.view(name='toalign')
reference_ligand.view(name='reflig')

#now we align the extracted ligand to the crystal of the reference
lig_aligned = maximalSubstructureAlignment(reference_ligand,ligand_to_align)
lig_aligned.view(name='shouldbealign')
```

I think that, in general, using boolean arrays instead of the indexes will break any correspondence between two sets of coordinates if they are not ordered in the same way (index 0 of ref corresponds to index 10 of query,  and index 2 of ref goes with index 5 of query, for example).

j3mdamas:
Check out #738. This fixes the issue, but maybe it's not the most elegant.

### DIFF
--- a/htmd/molecule/graphalignment.py
+++ b/htmd/molecule/graphalignment.py
@@ -5,6 +5,8 @@
 #
 import networkx as nx
 import numpy as np
+from unittest import TestCase
+import os
 
 
 def createProductGraph(G, H, tolerance, fields):
@@ -121,8 +123,8 @@ def maximalSubstructureAlignment(mol1, mol2, sel1='all', sel2='all', fields=('el
 
     _, _, matching = compareGraphs(g1, g2, fields=fields, tolerance=tolerance, returnmatching=True)
 
-    matchnodes1 = [x[0] for x in matching]
-    matchnodes2 = [x[1] for x in matching]
+    matchnodes1 = np.array([x[0] for x in matching])
+    matchnodes2 = np.array([x[1] for x in matching])
 
     mol2.align(sel=matchnodes2, refmol=mol1, refsel=matchnodes1)
 
@@ -134,3 +136,29 @@ def maximalSubstructureAlignment(mol1, mol2, sel1='all', sel2='all', fields=('el
         mol2.view(sel='all', style='Lines')
 
     return mol2
+
+
+class TestGraphAlignment(TestCase):
+    def test_maximalSubstructureAlignment(self):
+        from htmd.home import home
+        from htmd.molecule.molecule import Molecule
+
+        ref_lig = Molecule('2Y02')
+        ref_lig.filter('resname WHJ')
+        ref_lig.filter('residue 1')
+
+        lig2align = Molecule('2Y03')
+        lig2align.filter('resname 5FW')
+        lig2align.filter('residue 0')
+
+        lig_aligned = maximalSubstructureAlignment(ref_lig, lig2align)
+
+        lig_reference = Molecule(os.path.join(home(dataDir='test-molecule-graphalignment'), 'lig_aligned.pdb'))
+
+        self.assertTrue(np.allclose(lig_aligned.coords, lig_reference.coords, rtol=1e-4),
+                        'maximalSubstructureAlignment produced different coords')
+
+if __name__ == "__main__":
+    import unittest
+
+    unittest.main(verbosity=2)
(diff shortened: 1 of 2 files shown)


htmd/molecule/molecule.py
```diff
@@ -457,11 +457,12 @@ def align(self, sel, refmol=None, refsel=None, frames=None):
         frames = np.array(frames)
         # if not isinstance(refmol, Molecule):
         # raise NameError('Reference molecule has to be a Molecule object')
-        sel = self.atomselect(sel)
-        refsel = refmol.atomselect(refsel)
+        sel = self.atomselect(sel, indexes=True)
+        refsel = refmol.atomselect(refsel, indexes=True)
         if (type(sel[0]) == bool) and (np.sum(sel) != np.sum(refsel)):
```
stefdoerr: You need to fix this check now

stefdoerr:
Normally the `mol.align` is documented as being used with atomselect strings which intrinsically don't respect order. If you do in VMD `atomselect index 13 4 0 5` it will return the atoms in increasing index order. So what you propose by passing explicitely sorted index arrays is undefined (but nice) behaviour. Maybe we should document it? It gets a bit confusing as we should warn that atomselections don't respect order.

htmd/molecule/molecule.py
j3mdamas: right, right

j3mdamas:
well, yes, that is indeed a problem. maybe @alejandrovr can give some more input on his tests.

regarding the projections that were failing, one was `metricrmsd`. my second commit fixed it, you can see the differences between them.

stefdoerr:
My point was that MetricRMSD is not supposed to be used with different atom orderings. It's meant to calculate the RMSD of trajectories with reference to the starting structure or a reference structure of the same system with same atom ordering.

I assume it's documented somewhere in HTMD this thing about atomselect ordering (? maybe in the atomselect tutorial?) but we should definitely make it clearer.

alejandrovr:
I am not sure if I have understood correctly, but my concern was not about the ordering of the indexes per se. As long as the pairing between the two sets of coordinates is kept, the order does not matter. I will try to test this new version!

stefdoerr:
When you do an atom selection in VMD it will return the indexes of the atoms sorted. So if you do `index 7 4 9 1` you will get back `1 4 7 9` same with selections like `resname MOL and name C7 H2 CA` it will return back the atom indexes in sorted order.
This is why my code ignored the matching of the two atom sets that you passed it because the matching is just the order in the two sets.

j3mdamas:
You may re-review it @stefdoerr

htmd/molecule/graphalignment.py
```diff
@@ -133,4 +135,30 @@ def maximalSubstructureAlignment(mol1, mol2, sel1='all', sel2='all', fields=('el
...
\ No newline at end of file
+    return mol2
+
+
+class TestGraphAlignment(TestCase):
+    def test_maximalSubstructureAlignment(self):
+        from htmd.home import home
+        from htmd.molecule.molecule import Molecule
+
+        ref_lig = Molecule('2Y02')
```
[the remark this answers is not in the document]
j3mdamas: just wanted to make them reproducible. I added the files, I'll make the change

PR #741, opened 2018-07-24
Membrane builder tutorial

stefdoerr:
I added a membrane builder tutorial to advertise this as well.

In the process I also fixed some issues:
a) psfgen `call` was not logging the `stderr` to the file
b) psfgen fails in some locales so calling it with `LC_ALL=C` is safer
c) removed unused files and code
d) renamed private functions to start with `_`
e) `listLipids()` now also prints the location of the database file which I also moved to the lipids folder.


stefdoerr:
@cuzzo87 whenever you have time review this tutorial

stefdoerr:
@cuzzo87 you "requested changes". You should approve it instead. You already made the changes :)

PR #743, opened 2018-07-24
made the removed existing sentinel message debug level

stefdoerr:
Closes this https://github.com/Acellera/htmd/issues/661


stefdoerr:
I won't let it go through travis, it's pointless. Merged. Thanks

PR #748, opened 2018-07-25
SmallMol fixes and small improvements:

j3mdamas:
- adds resolution argument to depict method
- changes totalcharge to a true property and removes it from list of properties
- small formatting and synthax improvements (variable overshadowing, return simplifications, docstrings and typos)


stefdoerr:
While you are at it could you make the function argument naming style consistent in smallmol? Not that the rest of HTMD is extremely consistent but using lowercase, camelcase and underscores in a single function/method is a bit too much.

cuzzo87:
@stefdoerr I will do a review of that. I will modify thse "chimera" by looking at htmd code. Thank you for the suggestion

stefdoerr:
If you look at the HTMD code take care hahaha. It's also a bit
inconsistent. Since smallmol is beta we can make it better than HTMD! :D

PR #753, opened 2018-08-02
Improvements to custom QM calculators

stefdoerr:
- [x] custom QM class accept objects of calculators since they need to created in many cases with different parameters.
- [x] Minimizer classes to swap minimizers in the QM calculators
- [ ] Don't know what to do about the dipole?

Take a look at it @j3mdamas  and tell me what you think


j3mdamas:
In due time

stefdoerr:
image

stefdoerr:
Is the time due @j3mdamas ?

PR #757, opened 2018-08-22
Smallmol get bonds warning

cuzzo87:
Modified two methods.

1) getBonds that will not crush if no bonds exists in the molecule (eg single atom molecule)
2) toMolecule will raise an error if no atoms coordinates exist

### DIFF
--- a/htmd/smallmol/smallmol.py
+++ b/htmd/smallmol/smallmol.py
@@ -1179,7 +1179,8 @@ class SmallMol(object):
         # checks if coordintates exist
         n_coords = self.getCoords().shape[0]
         if n_coords == 0:
-            raise ValueError('No coordinates found. You need at least one atom coordinate.')
+            raise ValueError('No coordinates found. You need at least one atom coordinate to convert SmallMol'
+                             ' to a Molecule object.')
 
         molHtmd = None
         for n in ids:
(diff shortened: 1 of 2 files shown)


htmd/smallmol/smallmol.py
```diff
@@ -1215,7 +1220,7 @@ def getBonds(self):
         bondtypes = []
         _mol = self.toRdkitMol()
         for bo in _mol.GetBonds():  # self._mol.GetBonds():
-            bonds.append([bo.GetBeginAtomIdx(), bo.GetEndAtomIdx()])
+            bonds.append([bo.GetBeginAtomIdx(), bo.GetEndAtom_Idx()])
```
stefdoerr: The method name changed? How come?

htmd/smallmol/smallmol.py
```diff
@@ -1176,6 +1176,11 @@ class NoConformerError(Exception):
         elif not isinstance(ids, list):
             raise ValueError('The argument ids should be a list of confomer ids')
 
+        # checks if coordintates exist
+        n_coords = self.getCoords().shape[0]
+        if n_coords == 0:
+            raise ValueError('No coordinates found. You need at least one atom coordinate.')
```
stefdoerr: `No coordinates found. You need at least one atom coordinate to convert SmallMol to a Molecule object.`
make it like this a tiny bit more descriptive.

PR #759, opened 2018-08-22
Modified opm method

cuzzo87:
url for retrieving pdb has changed in opm webserver.
Using `4eiy` as example it has just modifeid from `http://opm.phar.umich.edu/pdb/4eiy.pdb` to `https://storage.googleapis.com/opm-assets/pdb/4eiy.pdb`


stefdoerr:
You already accidentally included this in the smallmol PR

PR #770, opened 2018-09-05
Update .lgtm.yml

samlanning:
Update the `.lgtm.yml` script to exclude javascript files from the result (now that there is no js in the project)

In response to https://discuss.lgtm.com/t/alerts-in-deleted-files/1138


stefdoerr:
Oh great! Thanks a lot for the PR! 😄

stefdoerr:
@samlanning because HTMD is in part commercialized by Acellera, we cannot accept a PR without a signed [contributors agreement](https://github.com/Acellera/htmd/blob/master/doc/source/developers/contributor_agreement.md) that allows them to commercialize it.

In practice if you don't care about these two lines of code I can just close the PR and write them myself into the file, then you don't need to bother with signing a document for two lines. As you prefer.

Thanks again for the fix and sorry for the bother.

samlanning:
@stefdoerr added the CLA :)

stefdoerr:
Thank you so much! :)

Issue #776, opened 2018-09-24
parameterization.writers error (PRM file) KeyError: ('C261', 'C261x4')


MBenabderrahmane:
Hi, I'm trying to use the parameterize module for my ligand and unfortunately I'm getting this error when the script try to write the final parameters 

== Writing results ==

2018-09-22 12:26:34,055 - htmd.parameterization.dihedral - INFO - Start parameter optimization
2018-09-22 12:26:34,055 - htmd.parameterization.dihedral - INFO - Initial RMSD: 5.099872 kcal/mol
2018-09-22 12:28:45,132 - htmd.parameterization.dihedral - INFO - Current RMSD: 0.515603 kcal/mol
2018-09-22 12:36:26,145 - htmd.parameterization.dihedral - INFO - Current RMSD: 0.490852 kcal/mol
2018-09-22 12:50:18,631 - htmd.parameterization.dihedral - INFO - Current RMSD: 0.407101 kcal/mol
2018-09-22 12:55:54,655 - htmd.parameterization.dihedral - INFO - Current RMSD: 0.257487 kcal/mol
2018-09-22 17:03:38,605 - htmd.parameterization.dihedral - INFO - Final RMSD: 0.257487 kcal/mol
2018-09-22 17:03:38,606 - htmd.parameterization.dihedral - INFO - Finished parameter optimization
2018-09-22 17:03:39,782 - htmd.parameterization.writers - INFO - Write MOL2 file: res/parameters/CGenFF_2b6/HF-6-31Gstar-vacuum/mol.mol2
2018-09-22 17:03:39,784 - htmd.parameterization.writers - INFO - Write PDB file: res/parameters/CGenFF_2b6/HF-6-31Gstar-vacuum/mol.pdb
2018-09-22 17:03:39,785 - htmd.parameterization.writers - INFO - Write COOR file: res/parameters/CGenFF_2b6/HF-6-31Gstar-vacuum/mol.coor
2018-09-22 17:03:39,787 - htmd.parameterization.writers - INFO - Write PSF file: res/parameters/CGenFF_2b6/HF-6-31Gstar-vacuum/mol.psf
2018-09-22 17:03:39,788 - htmd.parameterization.writers - INFO - Write RTF file: res/parameters/CGenFF_2b6/HF-6-31Gstar-vacuum/mol.rtf
Traceback (most recent call last):
  File "/home/med/.conda/envs/py35/bin/parameterize", line 11, in <module>
    sys.exit(main_parameterize())
[...]

PR #777, opened 2018-09-25
Fix PRM writer for CGENFF

### LINKED ISSUE #754: parametrize CGenFF
I am using parametrize with CGenFF. QM calculation successfully finishes, but parametrize fails while writing topologies and parameters files. The error is: 

```pytb
== Writing results ==

Traceback (most recent call last):
  File "/opt/miniconda3/bin/parameterize", line 11, in <module>
    sys.exit(main_parameterize())
  File "/opt/miniconda3/lib/python3.6/site-packages/htmd/parameterization/cli.py", line 288, in main_parameterize
    writeParameters(mol, parameters, qm, method, netcharge, args.outdir, original_coords=orig_coor)
  File "/opt/miniconda3/lib/python3.6/site-packages/htmd/parameterization/writers.py", line 312, in writeParameters
    writePRM(mol, parameters, file_)
  File "/opt/miniconda3/lib/python3.6/site-packages/htmd/parameterization/writers.py", line 154, in writePRM
    val = parameters.bond_types[type]
KeyError: ('C251x0',   'C251x1')
```

It seems that the error arises due to the fact that in the list returned by getSortedAndUniqueTypes the order of the atomtypes within each bond_type tuple may in some cases differ from those stored in the "parameters" object. (the same for angles, dihedrals, impropers).
For example in my case ('C251x0', 'C251x1') key is not found within parameters.bond_types, because it was stored as ('C251x1', 'C251x0')

It seems that with the following changes to getSortedAndUniqueTypes function it works (at least in my case):

```python
def getSortedAndUniqueTypes(types, field):
    if field == 'atom_types':
        return np.unique(types)
    #if field == 'bond_types':
    #     for i in range(types.shape[0]):
    #       types[i] = sorted(types[i])
    #elif field == 'angle_types':
    #     for i in range(types.shape[0]):
    #       types[i][0], types[i][2] = sorted([types[i][0], types[i][2]])
    #elif field == 'dihedral_types':
    #     for i in range(types.shape[0]):
    #        if types[i][0] > types[i][3]:
    #            types[i] = types[i][::-1]
    #elif field == 'improper_types' or field == 'improper_periodic_types':
    #    for i in range(types.shape[0]):
    #        types[i][0], types[i][1], types[i][3] = sorted([types[i][0], types[i][1], types[i][3]])
    #else:
    #    raise RuntimeError('Invalid field')
    return sorted(list({tuple(row) for row in types.tolist()}))
```

stefdoerr:
Closes #754  #776 

It was an issue where I didn't update the PRM writer when I made the `getParameters` function for the new parmed versions and it went unnoticed.

Since I was making changes I also made parameterize only print equivalents when they exist, instead of printing a list of all atoms of the molecule which is useless.

### DIFF
--- a/htmd/parameterization/cli.py
+++ b/htmd/parameterization/cli.py
@@ -111,7 +111,8 @@ def printReport(mol, netcharge, equivalents, all_dihedrals):
 
     print('Equivalent atom groups:')
     for atom_group in equivalents[0]:
-        print('  ' + ', '.join(mol.name[list(atom_group)]))
+        if len(atom_group) > 1:
+            print('  ' + ', '.join(mol.name[list(atom_group)]))
 
     print('Parameterizable dihedral angles:')
     for equivalent_dihedrals in all_dihedrals:
--- a/htmd/parameterization/writers.py
+++ b/htmd/parameterization/writers.py
@@ -151,19 +151,19 @@ def writePRM(mol, parameters, filename):
     print("BONDS", file=f)
     types = getSortedAndUniqueTypes(mol.atomtype[mol.bonds], 'bond_types')
     for type in types:
-        val = parameters.bond_types[type]
+        val = getParameter(type, parameters.bond_types)
         print("%-6s %-6s %8.2f %8.4f" % (type[0], type[1], val.k, val.req), file=f)
 
     print("\nANGLES", file=f)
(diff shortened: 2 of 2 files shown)


raimis:
@stefdoerr how come none of the tests haven't caught this?

PR #786, opened 2018-10-05
Refactor QMML

raimis:
- [x] Add `--qmml` flag
- [x]  Fix QMML set up
- [ ] Error checks and logging


raimis:
This is not actually merged! Need to be redone to keep GitHub happy...

PR #790, opened 2018-10-09
Fix Molecule align

stefdoerr:
I don't know how this bug went unnoticed but aligning on a specific frame zeroed all other frames. Anyway, I also added now an option to align all frames to all of the reference

### DIFF
--- a/htmd/molecule/molecule.py
+++ b/htmd/molecule/molecule.py
@@ -2071,7 +2071,7 @@ class TestMolecule(TestCase):
 
         refcoords = np.load(path.join(home(dataDir='test-molecule'), 'test-selfalign-mol.npy'))
 
-        assert np.allclose(mol.coords, refcoords, atol=1E-6)
+        assert np.allclose(mol.coords, refcoords, atol=1E-3)
 
     def test_alignToReference(self):
         from htmd.home import home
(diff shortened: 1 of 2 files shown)


htmd/molecule/molecule.py
```diff
@@ -488,19 +495,26 @@ def align(self, sel, refmol=None, refsel=None, frames=None):
         from htmd.molecule.util import _pp_align
         if refmol is None:
             refmol = self
+            if matchingframes:
+                raise ValueError('You cannot align a molecule\'s frames to themselves. '
+                                 'If you want to use the matchinframes option supply a reference molecule.')
```
j3mdamas: typo -> matchingframes

stefdoerr:
We should but currently I have not time. We can leave the PR for a moment I guess

j3mdamas:
ok. for me it looks good other than that.

htmd/molecule/molecule.py
```diff
@@ -1955,9 +1955,17 @@ def __init__(self, sel=None, style=None, color=None):
...
 class TestMolecule(TestCase):
-    def test_trajReadingAppending(self):
+    def setUp(self):
+        super().setUp()
         from htmd.home import home
 
+        self.trajmol = Molecule(path.join(home(dataDir='metricdistance'), 'filtered.pdb'))
+        self.trajmol.read(path.join(home(dataDir='metricdistance'), 'traj.xtc'))
+
+        self.mol3PTB = Molecule('3PTB')
```
j3mdamas: As a note for the future, we should be using `subTest` when working on the same object, to avoid loading the Molecule over and over again. It's not needed now.

htmd/molecule/molecule.py
```diff
@@ -2040,6 +2059,55 @@ def test_uniqueResidueID(self):
...
+        _ = mol2.filter('noh') # Remove some atoms to check aligning molecules with different numAtoms
+
+        mol.align('noh', refmol=mol2)
+
+        refcoords = np.load(path.join(home(dataDir='test-molecule'), 'test-align-refmol.npy'))
+
+        assert np.allclose(mol.coords, refcoords, atol=1E-6)
+        assert np.allclose(mol.coords[mol.atomselect('noh'), :, 3], mol2.coords[:, :, 0], atol=1E-6)
+
+    def test_alignToReferenceMatchingFrames(self):
```
j3mdamas: The `matchingframes` is a new argument, I thought we should be testing `frames` as well

stefdoerr: Done

PR #792, opened 2018-10-18
Fix Gasteiger charges

### LINKED ISSUE #789: fitGasteigerCharges breaks on a molecule
[mymol.zip](https://github.com/Acellera/htmd/files/2456790/mymol.zip)

```python
from htmd.charge import fitGasteigerCharges
from htmd.molecule.molecule import Molecule
mol = Molecule('/tmp/mymol.mol2')
fitGasteigerCharges(mol)
```

```pytb
/shared/sdoerr/Work/htmd/htmd/charge/charge.py in fitGasteigerCharges(mol)
     63 
     64     # Compute and store Gasteiger charges
---> 65     ComputeGasteigerCharges(rdkit_mol, throwOnParamFailure=True)
     66     mol = mol.copy()
     67     mol.charge[:] = [atom.GetDoubleProp('_GasteigerCharge') for atom in rdkit_mol.GetAtoms()]

ValueError: ERROR: No Gasteiger Partial Charge parameters for Element: C Mode: 
```

raimis:
- [x] Implement `fitChargesWithAntechamber` (not used now, but it is left for AM1-BCC option)
- [x] Guess bond type if needed
  - [x] Implement `guessBontType`
- [x] Fix #789
- [x] Update tests

### DIFF
--- a/htmd/charge/charge.py
+++ b/htmd/charge/charge.py
@@ -5,6 +5,7 @@
 #
 import logging
 import os
+import subprocess
 from tempfile import TemporaryDirectory
 
 import numpy as np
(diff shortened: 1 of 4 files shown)


stefdoerr:
Make it return any antechamber charges no? I'm using BCC currently

raimis:
Yes, it can compute BCC! Although, BCC charges for water molecule look a bit suspicious in comparison with ESP and Gasteiger.

htmd/charge/charge.py
```diff
@@ -69,6 +70,92 @@ def fitGasteigerCharges(mol):
     return mol
 
 
+def fitChargesWithAntechamber(mol, type='gas', molCharge=None):
```
stefdoerr: Are you using this somewhere? I could not find the reference.

htmd/charge/charge.py
```diff
@@ -69,6 +70,92 @@ def fitGasteigerCharges(mol):
...
+        raise TypeError('"mol" has to be instance of {}'.format(Molecule))
+    if mol.numFrames != 1:
+        raise ValueError('"mol" can have just one frame, but it has {}'.format(mol.numFrames))
+
+    if type not in ('gas', 'bcc'):
+        raise ValueError('"type" has to be "gas" or "bcc"')
+
+    if molCharge is None:
+        molCharge = int(round(np.sum(mol.charge)))
+        logger.info('Using partial atomic charges to calculate molecular charge')
```
stefdoerr: You could also report the total charge here maybe? Could be useful for detecting issues. Although I think cli.py is already logging the charges?

htmd/charge/charge.py
stefdoerr: Sorry, didn't see the comment above. Could you just add it as a BCC option for now? I need this to actually test my molecule dataset against parameterize

htmd/charge/charge.py
raimis: Yes, CLI already reports

htmd/charge/charge.py
raimis: AM1-BCC will be another PR. I also need implement symmetrization for them

PR #788, opened 2018-10-08
Fix molecule and atom renaming

raimis:
- [x] Enforce unique atom names
- [x] Clean up element guessing
- [x] Move atom renaming to `fftype` internals
  - [x] Update documentation
- [x] Update tests
- [x] Fix #524 
- [x] Fix #557
- [x] Fix #643 

This is a remake of #781 

Depend on https://github.com/Acellera/htmd-data/pull/10

PR #793, opened 2018-10-19
Make atom names unique if needed

raimis:
- [x] Make atom names unique if needed
  - [x] Implement `makeAtomNamesUnique`
- [x] Update tests

~~Depend on #788 and #792~~

Depend on https://github.com/Acellera/htmd-data/pull/11

### DIFF
--- a/htmd/parameterization/cli.py
+++ b/htmd/parameterization/cli.py
@@ -67,7 +67,7 @@ def getArgumentParser():
 def _prepare_molecule(args):
 
     from htmd.molecule.molecule import Molecule
-    from htmd.parameterization.util import guessElements
+    from htmd.parameterization.util import makeAtomNamesUnique, guessElements
 
     mol = Molecule(args.filename, guessNE=['bonds'], guess=[])
 
@@ -75,9 +75,14 @@ def _prepare_molecule(args):
     if mol.numFrames != 1:
         raise RuntimeError('{} has to contain only one molecule, but found {}'.format(args.filename, mol.numFrames))
 
-    # Check if each atom name is unique
+    # Make atom names unique if needed
     if np.unique(mol.name).size != mol.numAtoms:
-        raise RuntimeError('The atom names in {} has to be unique!'.format(args.filename))
+        logger.warning('Atom names in the molecule are not unique!')
+        new_mol = makeAtomNamesUnique(mol)
+        for i, (old_name, new_name) in enumerate(zip(mol.name, new_mol.name)):
+            if old_name != new_name:
+                logger.warning('Rename atom {:3d}: {:4s} --> {:4s}'.format(i, old_name, new_name))
+        mol = new_mol
 
     # Guess elements
     # TODO: it should not depend on FF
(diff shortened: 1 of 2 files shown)


j3mdamas:
I'm puzzled by this. What does this new function is supposed to do
different from the canolicalizeNames function?

htmd/parameterization/fftype.py
```diff
@@ -125,124 +142,114 @@ def fftype(mol, rtfFile=None, prmFile=None, method='GAFF2', acCharges=None, tmpD
                     raise RuntimeError('"antechamber" failed')
 
                 # Run parmchk2
-                returncode = subprocess.call([parmchk2_binary,
-                                              '-f', 'prepi',
-                                              '-s', atomtype,
-                                              '-i', 'mol.prepi',
-                                              '-o', 'mol.frcmod',
-                                              '-a', 'Y'], cwd=tmpdir)
+                cmd = ['parmchk2',
```
stefdoerr: Why this? I think it's better to check for the binary as it was before.

stefdoerr:
So what is the difference with canonicalizeAtomNames? That it maintains existing atom names? How did the previous fail? Would be useful I think to have a description in the PR

htmd/parameterization/cli.py
```diff
@@ -73,17 +73,22 @@ def getArgumentParser():
...
 
-    # Check if each atom name is unique
+    # Make atom names unique if needed
     if np.unique(mol.name).size != mol.numAtoms:
-        raise RuntimeError('The atom names in {} has to be unique!'.format(args.filename))
+        logger.warning('Atom names in the molecule are not unique!')
+        new_mol = makeAtomNamesUnique(mol)
+        for i, (old_name, new_name) in enumerate(zip(mol.name, new_mol.name)):
+            if old_name != new_name:
+                logger.warning('Rename atom {:3d}: {:4s} --> {:4s}'.format(i, old_name, new_name))
```
j3mdamas:
```suggestion
                logger.warning('Renamed atom {:3d}: {:4s} --> {:4s}'.format(i, old_name, new_name))
```

htmd/parameterization/util.py
```diff
@@ -296,6 +296,77 @@ def guessBondType(mol):
...
+    >>> new_mol = makeAtomNamesUnique(mol)
+    >>> assert new_mol is not mol
+    >>> new_mol.name
+    array(['1A', '1A0', 'A1B1', 'A1B2'], dtype=object)
+    """
+
+    from htmd.molecule.molecule import Molecule
+
+    if not isinstance(mol, Molecule):
+        raise TypeError('"mol" has to be instance of {}'.format(Molecule))
```
j3mdamas:
```suggestion
        raise TypeError('"mol" has to be an instance of {}'.format(Molecule))
```

htmd/parameterization/fftype.py
raimis: `subprocess.call` it does implicitly

stefdoerr: It was more to give a more sensible error message to the user but whatever. It's not supposed to be missing anyway

PR #803, opened 2018-11-07
Increases bounds for offset and checks if it's hit

stefdoerr:
Data updated: https://github.com/Acellera/htmd-data/pull/18


htmd/parameterization/dihedral.py
```diff
@@ -395,6 +395,10 @@ def _fit(self):
         finish = time.clock()
         logger.info('Finished parameter optimization after %f s' % (finish-start))
 
+        upper_bounds, lower_bounds = self._getBounds()
+        if vector[-1] == upper_bounds[-1] or vector[-1] == lower_bounds[-1]:
```
raimis: np.close(vector[-1], upper_bounds[-1], atol=0.01)

stefdoerr: done

PR #804, opened 2018-11-09
Update voxeldescriptors.py

hawk31:
More general element mapping

### DIFF
--- a/htmd/molecule/voxeldescriptors.py
+++ b/htmd/molecule/voxeldescriptors.py
@@ -259,7 +259,7 @@ def _getRadii(mol):
         'BR': 'Br'
     }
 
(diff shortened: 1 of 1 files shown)


htmd/molecule/voxeldescriptors.py
```diff
@@ -259,7 +259,7 @@ def _getRadii(mol):
         'BR': 'Br'
     }
 
-    for el in ['H', 'C', 'N', 'O', 'F', 'Mg', 'P', 'S', 'Cl', 'Ca', 'Fe', 'Zn', 'Br', 'I']:
+    for el in vdw.elements:
```
j3mdamas: what are you trying to do? vdw does not have an `elements`

PR #805, opened 2018-11-12
Update voxeldescriptors.py

hawk31:
Made the (hopefully) right modification now.

### DIFF
--- a/htmd/molecule/voxeldescriptors.py
+++ b/htmd/molecule/voxeldescriptors.py
@@ -259,7 +259,7 @@ def _getRadii(mol):
         'BR': 'Br'
     }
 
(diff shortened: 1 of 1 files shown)


j3mdamas:
Fine by me, but the way you changed now will make this if-clause https://github.com/Acellera/htmd/pull/805/files#diff-730a4afe430346fb62db4d7b944db4feR272 obsolete. But it's fine, there is more important maintenance to do than that :)

hawk31:
I did not notice that part :stuck_out_tongue: . I can make another PR with that removed.

j3mdamas:
don't worry about it. I think it's better to do these things after codes break apart

PR #807, opened 2018-11-13
Scipy minimizer

stefdoerr:
Changed minimizer to scipy.
Added tests for glycol since h2o2 doesn't have any atoms to minimize on the dihedral scans.
Reinitiate the MM minimizer with the better charges after the charge guessing.

Data PR: https://github.com/Acellera/htmd-data/pull/26


stefdoerr:
Done, uses forces now correctly. Fixed the two tests so that they pass travis

j3mdamas:
Anything I can do to help here? Or is it just solving the htmd/data conflict?

stefdoerr:
No, it's probably impossible to pass the tests until there is a way to disable dihedral fitting in parameterize. The numerical noise sums up between the different optimizations that in the end you get different albeit functionally equivalent parameterizations.

htmd/parameterization/cli.py
```diff
@@ -614,11 +607,15 @@ def main_parameterize(arguments=None):
         # Detect chiral centers
         initial_chiral_centers = detectChiralCenters(mol)
 
+        mm_minimizer = None
+        if args.min_type == 'mm' or args.dihed_opt_type == 'mm':
```
raimis: Why do you need the second condition? Isn't this enough:
```python
if args.min_type == 'mm':
```

htmd/parameterization/cli.py
```diff
@@ -628,6 +625,12 @@ def main_parameterize(arguments=None):
     # Fit charges
     mol = _fit_charges(mol, args, qm_calculator)
 
+    # Recreate MM minimizer once charges have been fitted
+    mm_minimizer = None
+    if args.min_type == 'mm' or args.dihed_opt_type == 'mm':
```
raimis: Why do you need the second condition? Isn't this enough:
```python
args.dihed_opt_type == 'mm':
```

htmd/parameterization/test.py
```diff
@@ -395,6 +395,18 @@ def test_h2o2_dihed_opt_mm_fake(self):
         self._run(refDir, resDir, 'parameterize input.mol2 -f GAFF2 --charge-type Gasteiger --min-type mm --scan-type mm --fake-qm')
         self._test(refDir, resDir)
 
+    def test_glycol_dihed_opt_mm_fake(self):
+        refDir = os.path.join(self.dataDir, 'glycol_dihed_opt_mm_fake')
+        resDir = os.path.join(self.testDir, 'glycol_dihed_opt_mm_fake')
+        self._run(refDir, resDir, 'parameterize input.mol2 -f GAFF2 --charge-type Gasteiger --min-type None --scan-type mm --fake-qm --dihed-num-searches 0')
+        self._test(refDir, resDir, energyProfileAbsTol=0.01)
```
raimis: How is it possible: energy needs a large tolerance, while the rest number pass with low tolerances

htmd/qm/custom.py
```diff
@@ -118,18 +114,27 @@ def minimize(self, coords, restrained_dihedrals=None):
...
+
+        def goalFunc(x):
+            sim.context.setPositions(x.reshape((natoms, 3)) * unit.angstrom)
+            state = sim.context.getState(getEnergy=True, getForces=True)
+            energy = state.getPotentialEnergy().value_in_unit(unit.kilocalories_per_mole)
+            forces = state.getForces(asNumpy=True).value_in_unit(unit.kilocalories_per_mole / unit.angstrom)
+            grad = -forces.reshape(-1)
+            return energy, grad
+
+        res = minimize(goalFunc, coords.reshape(-1), method='L-BFGS-B', jac=True, options={'ftol': 0, 'gtol': 0.05})
```
raimis: How have you the selected tolerances. It would be good to document with a comment.

htmd/parameterization/cli.py
```diff
@@ -628,6 +625,12 @@ def main_parameterize(arguments=None):
     # Fit charges
     mol = _fit_charges(mol, args, qm_calculator)
 
+    # Recreate MM minimizer once charges have been fitted
```
raimis: All this block should be inside:
```python
    if len(selected_dihedrals) > 0:
```

htmd/parameterization/cli.py
stefdoerr: No of course not. You could QM minimize the initial conformer and MM minimize the dihedrals

htmd/parameterization/cli.py
stefdoerr: See above answer

htmd/qm/custom.py
stefdoerr: https://wiki.fysik.dtu.dk/ase/ase/optimize.html
I copied the tolerance from the ASE tutorial which is as arbitrary as the other thresholds we choose I would say. We don't document the 1e-5 tolerance for fitting either. As to why force tolerance, scipy doesn't support absolute energy tolerances, and it makes sense to use force tolerance for minimization.

htmd/parameterization/cli.py
stefdoerr: Ok! moved :)

htmd/qm/custom.py
stefdoerr: I tried also the relative 1 kjoule/mol used in OpenMM but it seemed to finish too early on my tests without reaching a good minimum.

htmd/parameterization/test.py
stefdoerr: There is no fitting done so I assume it will be pretty exact. Anyway, I actually think 0.01 should be default on the tests since the files are written with that precision and any kind of noise could flip numbers that are on the 0.005 limit

htmd/parameterization/cli.py
stefdoerr: `min_type` is the initial minimization of the structure. `dihed_opt_type` is the dihedral scan minimization. In either of these cases you need a mm minimizer. Hope this explains it a bit clearer.

stefdoerr:
Answered the comments

htmd/parameterization/cli.py
stefdoerr: Aaaaah ok I see what you mean. Yes now that it's moved to the other section it makes sense

htmd/parameterization/cli.py
stefdoerr: Changed

htmd/parameterization/cli.py
stefdoerr: In the past it was only instantiated once. Now of course since it's reinstantiated I can change the ifs

htmd/parameterization/test.py
raimis: Which files do you refer to?

stefdoerr: The .dat files of the plots. Ah sorry ok I missed a zero it seems. So 0.001 should be the tolerance in my opinion.

stefdoerr: 0.002 actually since it needs to be lower-than

stefdoerr: Changed it to 0.002 now.

raimis: OK!

htmd/qm/custom.py
raimis: In what units is `gtol`? kcal/mol/A or kJ/mol/A?

stefdoerr:
kcal/mol/A as is written 3 lines above
`state.getForces(asNumpy=True).value_in_unit(unit.kilocalories_per_mole /
unit.angstrom)`

htmd/qm/custom.py
raimis: OK! It is tighter than Psi4 max force tolerance.

stefdoerr:
I have no clue how the second dihedral skipped getting added :man_shrugging: . Added now and rerunning the tests

raimis:
Thanks @stefdoerr!

PR #810, opened 2018-11-15
Chiral center checks

raimis:
- [ ] Implement chiral center detection
- [ ] Enable chiral center check during optimizations

### DIFF
--- a/htmd/parameterization/util.py
+++ b/htmd/parameterization/util.py
@@ -383,6 +383,60 @@ def makeAtomNamesUnique(mol):
 
     return mol
 
+def detectChrilCenters(mol):
+    """
+    Detect chiral centers
+
+    Parameters
+    ----------
+    mol: Molecule
+        Molecule to detect chiral centers
+
+    Return
+    ------
+    results: list of tuples
+        List of chiral centers as tuples made of an atom index and a label ('S', 'R', '?')
+
+    Examples
+    --------
+    >>> from htmd.home import home
+    >>> from htmd.molecule.molecule import Molecule
+
+    >>> molFile = os.path.join(home('test-param'), 'H2O2.mol2')
+    >>> mol = Molecule(molFile)
+    >>> detectChrilCenters(mol)
+    []
+
+    >>> molFile = os.path.join(home('test-param'), 'fluorchlorcyclopronol.mol2')
+    >>> mol = Molecule(molFile)
+    >>> detectChrilCenters(mol)
+    [(0, 'R'), (2, 'S'), (4, 'R')]
+    """
+
+    from htmd.molecule.molecule import Molecule
+    from rdkit.Chem import MolFromMol2File, AssignAtomChiralTagsFromStructure, FindMolChiralCenters
+
+    if not isinstance(mol, Molecule):
+        raise TypeError('"mol" has to be instance of {}'.format(Molecule))
+    if mol.numFrames != 1:
+        raise ValueError('"mol" can have just one frame, but it has {}'.format(mol.numFrames))
+
+    # Set atom types to elements, overwise rdkit refuse to read a MOL2 file
+    htmd_mol = mol.copy()
+    htmd_mol.atomtype = htmd_mol.element
+
+    # Convert Molecule to rdkit Mol
+    with TemporaryDirectory() as tmpDir:
+        filename = os.path.join(tmpDir, 'mol.mol2')
+        htmd_mol.write(filename)
+        rdkit_mol = MolFromMol2File(filename, removeHs=False)
+    assert mol.numAtoms == rdkit_mol.GetNumAtoms()
+
+    # Detect chiral centers and assign their labels
+    AssignAtomChiralTagsFromStructure(rdkit_mol)
+    chiral_centers = FindMolChiralCenters(rdkit_mol, includeUnassigned=True)
+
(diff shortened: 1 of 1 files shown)


raimis:
@stefdoerr could you review, please.

htmd/parameterization/util.py
```diff
@@ -383,6 +383,72 @@ def makeAtomNamesUnique(mol):
...
+    """
+
+    from htmd.molecule.molecule import Molecule
+    from rdkit.Chem import MolFromMol2File, AssignAtomChiralTagsFromStructure, FindMolChiralCenters
+
+    if not isinstance(mol, Molecule):
+        raise TypeError('"mol" has to be instance of {}'.format(Molecule))
+
+    all_chiral_centers = []
+
```
stefdoerr: Copy mol once here, set atomtypes once. `htmd_mol = mol.copy()  htmd_mol.atomtype = htmd_mol.element`
In the loop just change the `htmd_mol.frame = i` variable and it will write the correct frame out

htmd/parameterization/util.py
```diff
@@ -383,6 +383,72 @@ def makeAtomNamesUnique(mol):
...
+            filename = os.path.join(tmpDir, 'mol.mol2')
+            htmd_mol.write(filename)
+            rdkit_mol = MolFromMol2File(filename, removeHs=False)
+        assert mol.numAtoms == rdkit_mol.GetNumAtoms()
+
+        # Detect chiral centers and assign their labels
+        AssignAtomChiralTagsFromStructure(rdkit_mol)
+        chiral_centers = FindMolChiralCenters(rdkit_mol, includeUnassigned=True)
+
+        all_chiral_centers.append(chiral_centers)
```
stefdoerr: Why would you want to check all frames of a molecule for chiral centers? I would rather make the function work on the first frame only like most molecule utils. From what I see you only use the first frame anyway.

htmd/parameterization/cli.py
```diff
@@ -106,6 +107,13 @@ def _prepare_molecule(args):
 
     # TODO: check bonds
 
+    # Detect chiral centers
+    mol.chiral_centers = detectChiralCenters(mol)[0]
```
stefdoerr: I'm not a huge fan of storing arbitrary values inside the Molecule class. Would it be much uglier to keep the chiral centers around as a variable?

htmd/parameterization/util.py
raimis: I agree, this could be simplified!

htmd/parameterization/cli.py
raimis: Cluttering the high-level code with low-level variables is also not good. On the other hand, chirallity is an intrinsic property of a molecule. What do you propose?

stefdoerr: What you are doing here is storing volatile data inside the class. It is a volatile property that depends on changes to the coordinates so it should not be stored, but calculated on request instead. I would be fine for example if there was a `mol.detectChiralCenters()` method but I would not store the data in the class. 

So the data should be passed around as a low-level variable yes, since it's only used to compare an old state of the molecule to a new state of the molecule.

stefdoerr: All other properties we store in the Molecule are non-volatile. You can argue that bonds and charges are volatile in QM but HTMD is built for MD so we assume that they are not and thus bonds, dihedrals, angles, charges etc can be stored inside the class without ever changing

raimis:
I'm trying to generalize such checks.

stefdoerr:
I like it, but I will make a modest proposal on rewriting the filtering function like this to 

1. remove a few loops and unneeded temporary lists
2. avoid typechecking of mol in each iteration
3. avoid recalculating initial chiral centers for each result (they don't change)
4. avoid copying molecule multiple times

What do you think?

```python
def filterQMResults(all_results, mol=None):
    if mol and not isinstance(mol, Molecule):
        raise TypeError('"mol" has to be instance of {}'.format(Molecule))
        initial_chiral_centers = detectChiralCenters(mol)
        mol = mol.copy()

    all_valid_results = []
    for results in all_results:

        valid_results = []
        for result in results:
            valid = True
            # Remove failed QM results
            if not isinstance(result, QMResult):
                raise TypeError('"results" has to a list of list of {} instance'.format(QMResult))
            if result.errored:
                logger.warning('Rotamer is removed due to a failed QM calculations')
                continue

            # Remove results with wrong chiral centers
            if mol:
                mol.coords = result.coords
                chiral_centers = detectChiralCenters(mol)
                if intial_chiral_centers != chiral_centers:
                    logger.warning('Rotamer is removed due to a change of chiral centers: '
                                   '{} --> {}'.format(initial_chiral_centers, chiral_centers))
                    continue

            valid_results.append(result)

        # Remove QM results with too high QM energies (>20 kcal/mol above the minimum)
        qm_min = np.min([result.energy for result in valid_results])
        valid_results = [result for result in valid_results if (result.energy - qm_min) < 20]

        if len(valid_results) < 13:
            raise RuntimeError('Less than 13 valid QM results per dihedral. Not enough to fit!')

        all_valid_results.append(valid_results)

    return all_valid_results
```

stefdoerr:
Aaaah sorry, typo

stefdoerr:
Ok fixed the typeerror

PR #815, opened 2018-11-21
Fit initial charge with Gasteiger

raimis:
- [x] Fit initial charge with Gasteiger
- [x] Add tests

~~Depends on #782~~
Depends on https://github.com/Acellera/htmd-data/pull/23

### DIFF
--- a/htmd/parameterization/cli.py
+++ b/htmd/parameterization/cli.py
@@ -339,10 +339,10 @@ def _fit_initial_charges(mol, args):
             logger.info('Initial atomic charges are not required')
 
         else:
-            raise ValueError()
+            raise AssertionError()
 
     else:
-        raise ValueError()
+        raise AssertionError()
 
     return mol
 
--- a/htmd/parameterization/cli.py
+++ b/htmd/parameterization/cli.py
@@ -306,6 +306,45 @@ def _get_initial_parameters(mol, args):
     return mol, parameters
 
 
+def _fit_initial_charges(mol, args):
+
+    from htmd.charge import fitGasteigerCharges
+    from htmd.parameterization.util import guessBondType
+
+    logger.info('=== Initial atomic charge fitting ===')
+
+    if args.charge_type == 'None':
+        logger.info('Initial atomic charges are taken from {}'.format(args.filename))
+
+    elif args.charge_type in ('Gasteiger', 'AM1-BCC', 'ESP'):
+        if args.min_type == 'mm':
+            logger.info('Method: Gasteiger')
+
+            # TODO move to _prepare_molecule
+            if np.any(mol.bondtype == "un"):
+                logger.info('Guessing bond types')
+                mol = guessBondType(mol)
+
+            mol = fitGasteigerCharges(mol)
+
+            # Print the initial charges
+            logger.info('Initial atomic charges:')
+            for name, charge in zip(mol.name, mol.charge):
+                logger.info('   {:4s}: {:6.3f}'.format(name, charge))
+            logger.info('Molecular charge: {:6.3f}'.format(np.sum(mol.charge)))
+
+        elif args.min_type in ('None', 'qm'):
+            logger.info('Initial atomic charges are not required')
+
+        else:
+            raise ValueError()
+
+    else:
+        raise ValueError()
+
+    return mol
+
+
 def _fit_charges(mol, args, qm):
 
     from htmd.charge import fitGasteigerCharges, fitChargesWithAntechamber, fitESPCharges, symmetrizeCharges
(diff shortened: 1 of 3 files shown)


raimis:
@stefdoerr could you review, please!

stefdoerr:
Why are you not using the _fit_charges method?

raimis:
Look more carefully into the logic!

stefdoerr:
I don't have time currently so a TL;DR would be nice to help understand the code

htmd/parameterization/cli.py
```diff
@@ -308,6 +308,45 @@ def _get_initial_parameters(mol, args):
...
+    logger.info('=== Initial atomic charge fitting ===')
+
+    if args.charge_type == 'None':
+        logger.info('Initial atomic charges are taken from {}'.format(args.filename))
+
+    elif args.charge_type in ('Gasteiger', 'AM1-BCC', 'ESP'):
+        if args.min_type == 'mm':
+            logger.info('Method: Gasteiger')
+
+            # TODO move to _prepare_molecule
```
stefdoerr: I don't like this code duplication really.
I would suggest that we don't pass `args` to functions but instead pass the specific args. This makes the functions much mode modular and usable from outside the CLI application. In the future I'd like to separate the `argparser` from the `main` function of parameterize anyway so there should be no `args` object, so that it can be used better programmatically. For example now to use `_fit_charges` with `Gasteiger` if you want to do `ESP` later it's a pain in the butt (which is why I assume you made this code). 

For my evaluation for example I had to create fake `args` objects just to pass arguments to the function.
```python
class ChargeArgs:
    def __init__(self, charge_type, charge):
        self.charge_type = charge_type
        self.fix_charge = []
        self.charge = charge
        self.seed = 20170920
        self.outdir = tempfile.TemporaryDirectory().name

fftypedmol = _fit_charges(fftypedmol, ChargeArgs('Gasteiger', totalcharge), None)
```

So I would propose this change. Make `_fit_charges(mol, args, qm)` to `_fit_charges(mol, charge_type, charge, fix_charge, qm=None, outdir=None, seed=None)`. Then here just call it with `_fit_charges(mol, 'Gasteiger', charge, fix_charge)` and this we avoid data duplication and go a step further in making the code more modular.

raimis: The underscore prefixed functions in `cli.py` are not supposed to be neither reusable nor modular. It is just a stages of the parameterization protocol.

The initial charges fitting and the charge fitting are stages with different flow control, computation, and log messages.

The testing should be done with the whole tools, as much as possible, not its ripped off components. Otherwise important issues, such a lack of initial charges, pass unnoticed.

stefdoerr: If we want to do unit testing we need to modularize the code.

With code duplication you introduce modes of failure when changing stuff. Like the bond guessing which is duplicated here for the Gasteiger prediction method.

stefdoerr: For example here you are also missing the checks that you have in the _fit_charges. Would you want to not do those checks?

stefdoerr: Hm, I'll need to think about it.

htmd/parameterization/cli.py
```diff
@@ -308,6 +308,45 @@ def _get_initial_parameters(mol, args):
...
+            logger.info('Initial atomic charges:')
+            for name, charge in zip(mol.name, mol.charge):
+                logger.info('   {:4s}: {:6.3f}'.format(name, charge))
+            logger.info('Molecular charge: {:6.3f}'.format(np.sum(mol.charge)))
+
+        elif args.min_type in ('None', 'qm'):
+            logger.info('Initial atomic charges are not required')
+
+        else:
+            raise ValueError()
```
stefdoerr: Ok anyway I'll accept the PR like this but what is with these two empty valuerrors? I assume they won't be reached due to the argparser but then why add them? If you keep them include some message

htmd/parameterization/cli.py
raimis: This is high-level code, so the unit tests are effectively the functional tests from `test.py`

Regarding the bond guessing, it was marked for moving into `_prepare_molecule`. I'm still cleaning up `cli.py`, but not in this PR.

htmd/parameterization/cli.py
raimis: Yes, these are not suppose to reached, just safeguards, if we change something in the future.

raimis: There is no such thing as `CodeConsistencyError`, I used `ValueError`

stefdoerr: It's called `AssertionError` or `assert` ;)

raimis: OK! I will change...

PR #816, opened 2018-11-22
Fix NNP (a.k.a. QMML)

raimis:
- [x] Rename QMML to NNP
- [x] Fix NNP configuration
  - [x] Factor out `_get_queue`
  - [x] Split _get_reference_calculator into `_get_qm_calculator` and `_get_nnp_calculator`
- [x] Simplify parameter directory

~~Depend on #815~~
Depend on https://github.com/Acellera/htmd-data/pull/24

### DIFF
--- a/htmd/parameterization/cli.py
+++ b/htmd/parameterization/cli.py
@@ -512,21 +512,20 @@ VdW      : {VDW_ENERGY}
         logger.info('   {:8s} : {:10.3f} kcal/mol'.format(name, energies[name]))
 
 
-def _output_results(mol, parameters, original_coords, qm, args):
+def _output_results(mol, parameters, original_coords, args):
 
-    from htmd.parameterization.util import _qm_method_name
     from htmd.parameterization.writers import writeParameters
 
     logger.info('=== Results ===')
 
     # Output the FF parameters and other files
     # TODO get rid of QM
-    paramoutdir = os.path.join(args.outdir, 'parameters', args.forcefield, _qm_method_name(qm))
+    dir = os.path.join(args.outdir, 'parameters')
     # TODO split into separate writer
-    writeParameters(paramoutdir, mol, parameters, args.forcefield, args.charge, original_coords=original_coords)
+    writeParameters(dir, mol, parameters, args.forcefield, args.charge, original_coords=original_coords)
 
     # Write energy file
-    energyFile = os.path.join(paramoutdir, 'energies.txt')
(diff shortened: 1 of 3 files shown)


raimis:
@stefdoerr can you review, please.

htmd/parameterization/cli.py
```diff
@@ -473,21 +509,20 @@ def _printEnergies(molecule, parameters, filename):
...
-def _output_results(mol, parameters, original_coords, qm, args):
+def _output_results(mol, parameters, original_coords, args):
 
-    from htmd.parameterization.util import _qm_method_name
     from htmd.parameterization.writers import writeParameters
 
     logger.info('=== Results ===')
 
     # Output the FF parameters and other files
     # TODO get rid of QM
```
stefdoerr: you can remove this comment now

htmd/parameterization/cli.py
```diff
@@ -606,20 +656,21 @@ def main_parameterize(arguments=None):
 
         # Set parameters for the fake QM
         if args.fake_qm:
-            qm._parameters = parameters
+            assert not args.nnp
```
stefdoerr: Remove the assertion. You already have one above so it will never trigger

htmd/parameterization/cli.py
```diff
@@ -525,13 +564,23 @@ def main_parameterize(arguments=None):
...
-    # Get a reference calculator
-    qm = _get_reference_calculator(args)
+    # Get a queue
+    queue = _get_queue(args)
+
+    # Get calculators
+    qm = _get_qm_calculator(args, queue)
+    nnp = _get_nnp_calculator(args, queue)
+    if args.nnp:
+        ref = nnp
```
stefdoerr: Call ref calculator instead (as the functions already call it). Ref is too abstract a variable name I think

stefdoerr:
Can you also remove the redundant assertion and the old comment so that we can merge this @raimis ?

htmd/parameterization/cli.py
raimis: Renamed!

htmd/parameterization/cli.py
raimis: Removed!

htmd/parameterization/cli.py
raimis: I prefer to keep it, to be safe in future refactoring.

raimis:
@stefdoerr any more comments?

PR #817, opened 2018-11-23
fixed conda-build version to the one that works

stefdoerr:


raimis:
OK! Let's if it works...

stefdoerr:
It seems to be working though since it got further than the first few minutes when it used to crash :)

j3mdamas:
perfecto

stefdoerr:
woohoo!

PR #821, opened 2018-11-23
[TEST] drop low force constant dihedral terms

raimis:


j3mdamas:
Why do you need to make a PR for a test? Travis builds on branches too...

raimis:
It is not the case, if the branch is on my fork.

PR #827, opened 2018-11-28
Hide irrelevant log messages

raimis:
- [x] Hide `antechmber` stdout and stderr
- [x] Hide `MATCH` stdout and stderr
- [x] Hide irrelevant renaming logs
- [x] Set up only required calculators

### DIFF
--- a/htmd/parameterization/cli.py
+++ b/htmd/parameterization/cli.py
@@ -294,28 +294,25 @@ def _get_qm_calculator(args, queue):
 
     return qm
 
+
 def _get_nnp_calculator(args, queue):
 
-    if args.nnp:
-        import importlib
-        from htmd.qm.custom import CustomQM
+    import importlib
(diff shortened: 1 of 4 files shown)


raimis:
@stefdoerr can you review, please!

htmd/parameterization/cli.py
```diff
@@ -571,9 +568,17 @@ def main_parameterize(arguments=None):
     # Get a queue
     queue = _get_queue(args)
 
-    # Get calculators
-    qm_calculator = _get_qm_calculator(args, queue)
-    nnp_calculator = _get_nnp_calculator(args, queue)
+    # Get QM calculators
+    qm_calculator = None
+    if args.min_type == 'qm' or args.charge_type == 'ESP' or (args.fit_dihedral and not args.nnp):
```
stefdoerr: Shouldn't you also check if the `scan_type` is `qm`?

PR #829, opened 2018-11-28
Refactor dihedral scanning, fitting, and plotting

raimis:
- [ ] Clean up
- [x] Split dihedral scanning and fitting
- [ ] Split dihedral fitting and plotting
- [ ] Improve plotting


stefdoerr:
I like the refactoring work but could it wait until publication? Every such merge messes up my evaluation if some bugfix or improvement comes after it that I need to use.

PR #831, opened 2018-11-30
Make parameterize tests to work in a container

raimis:
- [x] Disable visualization files in ESP tests
- [x] Write the temporary files to `/tmp` only

### DIFF
--- a/htmd/charge/esp.py
+++ b/htmd/charge/esp.py
@@ -64,7 +64,7 @@ class MoleculeGrid:
     >>> from htmd.molecule.molecule import Molecule
     >>> molFile = os.path.join(home('test-charge'), 'H2O.mol2')
     >>> mol = Molecule(molFile, guessNE='bonds', guess=('angles', 'dihedrals'))
-    >>> #mol.write('H2O.xyz')  # Uncomment for visualization
+    >>> mol.write('H2O.xyz') # doctest: +SKIP
 
     >>> np.random.seed(20181113)
     >>> grid = MoleculeGrid(mol)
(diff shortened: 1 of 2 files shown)


raimis:
@stefdoerr can you review, please!

htmd/charge/esp.py
```diff
@@ -92,7 +92,7 @@ class MoleculeGrid:
            [ 1.35323883, -3.08583052, -0.17044048],
            [-0.80667116,  0.64144373, -0.99312377],
            [ 2.45040501, -2.23107494, -0.48419202]])
-    >>> grid.writeXYZ('H2O_1_2__50.xyz')
+    >>> #grid.writeXYZ('H2O_1_2__50.xyz')  # Uncomment for visualization
```
stefdoerr: What do you mean by this comment? Visualization? It's file writing

raimis: Visual inspection of the molecular grid with VMD.

stefdoerr: So the issue is that you cannot write in cwd in the container?

raimis: Yes!

stefdoerr: Could you try maybe stuff like 2) in this solution? https://stackoverflow.com/a/47362117
It's possible to change the test dir for doctests and I'd prefer that if you want to keep the doctests alive

raimis: The test are alive, just a files for visualization are not made.

raimis:
@stefdoerr I have skipped the lines in the doctests. Any more comments?

PR #832, opened 2018-12-04
Fixes conda-build

j3mdamas:
`conda build` of htmd was no longer working anymore with the latest version of `conda-build`. This fixes it. It's urgent and necessary to be merged on other branches as well.

### DIFF
--- a/package/htmd/build.sh
+++ b/package/htmd/build.sh
@@ -39,6 +39,9 @@ else
     exit 1
 fi
 
+# Delete compiled libs from other OS
+find $DIR/htmd/lib -maxdepth 1 ! -name "$OSNAME" ! -path $DIR/htmd/lib -type d -exec rm -rf {} \; -print || true
+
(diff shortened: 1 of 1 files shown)


package/htmd/build.sh
```diff
@@ -39,6 +39,9 @@ else
     exit 1
 fi
 
+# Delete compiled libs from other OS
+find $DIR/htmd/lib -maxdepth 1 ! -name "$OSNAME" ! -path $DIR/htmd/lib -type d -exec rm -rf {} \; -print || true
```
stefdoerr: I don't get the `! -path $DIR/htmd/lib` part of the command

j3mdamas: find gives you always the source folder of search (`$DIR/htmd/lib` in this case). It's just a way of excluding it

stefdoerr: merci, didn't know that

PR #833, opened 2018-12-07
Details which matter...

raimis:
... or why we need NNP tests!


raimis:
@stefdoerr could you review, please!

stefdoerr:
I only see the fix from npp to nnp. Did you really want a review for that? :P

PR #840, opened 2019-01-15
Parameterize: improve logging of dihedral scanning and parameter fitting

raimis:
- [x] Reorder atom type creation
- [x] Improve log of scanning
- [x] Improve log of fitting

### DIFF
--- a/htmd/parameterization/util.py
+++ b/htmd/parameterization/util.py
@@ -145,10 +145,14 @@ def scanDihedrals(mol, ref, dihedrals, outdir, scan_type='qm', mm_minimizer=None
     """
     num_rotamers = 36  # Number of rotamers for each dihedral to compute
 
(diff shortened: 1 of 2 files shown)


raimis:
@stefdoerr could you review this, please.

htmd/parameterization/util.py
```diff
@@ -180,6 +186,7 @@ def scanDihedrals(mol, ref, dihedrals, outdir, scan_type='qm', mm_minimizer=None
         directories.append(directory)
 
     # Setup and submit QM calculations
+    logger.info('Compute rotamer energies for:')
```
stefdoerr: Why not before the next loop where the rest of the printing is happening?

raimis: I moved them together.

PR #862, opened 2019-03-27
Fix the failure of chiral centre detection

raimis:
- [x] Preserve initial molecule
- [x] Make the chiral centre detector to accept custom atom types
- [x] Pass the initial atom types to the chiral centre detector

### DIFF
--- a/htmd/parameterization/cli.py
+++ b/htmd/parameterization/cli.py
@@ -578,8 +578,8 @@ def main_parameterize(arguments=None):
     # Get a molecule and check its validity
     mol = _prepare_molecule(args)
 
-    # Copy the molecule to preserve initial coordinates
-    initial_coords = mol.coords.copy()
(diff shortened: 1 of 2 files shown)


raimis:
@cuzzo87 Here is the fix for you! Check and review it.

cuzzo87:
For the error I was mentioning this fix should solve the problem. Anyway, did you try to load/parameterize the output.mol2 of parameterize ? If I recall correct, the reason for having the `atomtypes=element` was because rdkit cannot read our atomtypes. Can you check this ?

raimis:
Yes, it will fail with wrong types. As I said, this makes the user responsible for correct atom types, because `atomtype=element` is not good for rdkit.

cuzzo87:
For me it is ok.

raimis:
@cuzzo87 you need to approve, so I can merge.

PR #867, opened 2019-04-26
prevent charge.py doctest erroring when HTMD_CONFIG is defined

j3mdamas:
This happens on deployment sites where `HTMD_CONFIG` is defined and the doctest errors because htmd import is no longer silent. Not urgent, but annoying.


j3mdamas:
The solution adopted was copied from `detect.py`

PR #871, opened 2019-05-06
ignore elements in protocol

stefdoerr:
ignore elements in protocol since amber.build sometimes creates weird atom elements


j3mdamas:
For me it's OK, but why? I mean, is the use case standard or something particular for what you are working on? I'm asking this because we may be fixing something on the wrong place (I don't have enough context to know)

stefdoerr:
Yeah actually I see now the problem. It's rather an issue with the PSF reader. My bad

PR #875, opened 2019-05-22
re-enable element validation.

stefdoerr:
the issue was on moleculekit guessing elements badly and is now fixed. My bad for having pushed this


stefdoerr:
For posterity, moleculekit checks validity of elements when reading a file. Before, for PSF and PRMTOP it was guessing elements by the atom names which is messy and lead to badly guessed elements so I needed to disable the element validation. Then I improved moleculekit readers to guess element by the masses for PSF and PRMTOP files which fixed the issue and now validation can be done again.

PR #879, opened 2019-05-28
Implements overriding of Amber default search paths

### LINKED ISSUE #594: Override tleap default search paths
Following this comment: https://github.com/Acellera/htmd/issues/590#issuecomment-373505206

I think the approach is:

- [x] use `teLeap` instead of `tleap`
- [x] create `teleapimports` argument which defaults to normal behavior of `tleap` but can be overridden by the user
- [x] create a test case

j3mdamas:
Closes #594 (details of what was done can be checked there).

Small fix on offlibraries loading.

### DIFF
--- a/htmd/builder/amber.py
+++ b/htmd/builder/amber.py
@@ -55,6 +55,13 @@ def defaultAmberHome(teleap=None):
     return os.path.normpath(os.path.join(os.path.dirname(teleap), '../'))
 
 
+_defaultAmberSearchPaths = {'ff': join('dat', 'leap', 'cmd'),
+                            'topo': join('dat', 'leap', 'prep'),
+                            'param': join('dat', 'leap', 'parm'),
+                            'lib': join('dat', 'leap', 'lib')
+                            }
+
+
 def htmdAmberHome():
     """ Returns the location of the AMBER files distributed with HTMD"""
 
@@ -79,7 +86,7 @@ def listFiles():
     amberhome = defaultAmberHome()
 
     # Original AMBER FFs
-    ffdir = join(amberhome, 'dat', 'leap', 'cmd')
+    ffdir = join(amberhome, _defaultAmberSearchPaths['ff'])
     ffs = glob(join(ffdir, '*'))
     print('---- Forcefield files list: ' + join(ffdir, '') + ' ----')
     for f in sorted(ffs, key=str.lower):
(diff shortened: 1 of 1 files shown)


htmd/builder/amber.py
```diff
@@ -406,26 +431,40 @@ def build(mol, ff=None, topo=None, param=None, prefix='structure', outdir='./bui
...
+            teleapimports += [os.path.join(htmdamberdir, os.path.dirname(f))
+                              for f in ff if os.path.isfile(os.path.join(htmdamberdir, f))]
+            if len(teleapimports) == 0:
+                raise RuntimeError('No default Amber force-field imports found. Check '
+                                   '`htmd.builder.amber.defaultAmberHome()` and `htmd.builder.amber.htmdAmberHome()`')
+        # Set import flags for teLeap
+        teleapimportflags = []
+        for p in teleapimports:
+            teleapimportflags.append('-I')
+            teleapimportflags.append('{}'.format(p))
```
stefdoerr: `'{}'.format(p)` seems excessive. Should be equivalent to `str(p)`

htmd/builder/amber.py
```diff
@@ -406,26 +431,40 @@ def build(mol, ff=None, topo=None, param=None, prefix='structure', outdir='./bui
...
-            extrasource.append('-I')
-            extrasource.append('{}'.format(p))
+        if not teleapimports:
+            teleapimports = []
+            # Source default Amber (i.e. the same paths tleap imports)
+            amberhome = defaultAmberHome(teleap=teleap)
+            teleapimports.append(os.path.join(amberhome, 'dat', 'leap', 'prep'))
+            teleapimports.append(os.path.join(amberhome, 'dat', 'leap', 'lib'))
+            teleapimports.append(os.path.join(amberhome, 'dat', 'leap', 'parm'))
+            teleapimports.append(os.path.join(amberhome, 'dat', 'leap', 'cmd'))
```
stefdoerr: Why not do these like in charmm.build where you have a `defaultFF` function that returns them?

htmd/builder/amber.py
j3mdamas: sure, I can do that!

htmd/builder/amber.py
j3mdamas: In Amber we also have the `defaultFF` functions. But they return files, while these are search paths. To be honest, I wanted to unify them and replace their use in `listFiles` and `_locateFile`, but I thought that it was already too much refactoring. But if you agree with it, I could create some private functions that return those default paths, perhaps in a dictionary.

stefdoerr: Yes, maybe do that if it's not a bother. It's a bit ugly like this.

j3mdamas: It's not, I wanted to do it, but maybe too much refactoring. Let me check it.

j3mdamas: I think I was able to do what you asked in a very simple and clean way. Let me know if you like it.

htmd/builder/amber.py
```diff
@@ -131,33 +138,14 @@ def listFiles():
 def _locateFile(fname, type, teleap):
```
stefdoerr: sorry, I'll have to ask for a tiny change :P `type` is a python keyword. Try not to use it as a variable (might have been me who did it first but anyway)

stefdoerr:
very nice :)

htmd/builder/amber.py
j3mdamas: We were thinking the same at the same time and already pushing a fix. Done.

stefdoerr:
When it passes we can merge :)

j3mdamas:
I'd like for @cuzzo87 to look at it, since he's responsible for `SystemBuilder`. So, you can approve, and when he approves, we can merge. Sounds good?

j3mdamas:
Thanks @stefdoerr. @cuzzo87, whenever you have some time, could you look at this? Thanks!

PR #886, opened 2019-06-07
Create a share folder for HTMD with all the static files in it

stefdoerr:
Before it was a mess where every module had it's own files lying around. This:
a) caused a mess with pmbuilder since you need to copy each file individually
b) caused issues since some methods search for files relative to the module path while others relied on htmd.home
c) lack of cleanliness and standardization

Now all data necessary for HTMD modules will be stored in the share folder.


htmd/builder/charmmfiles/.gitignore
```diff
@@ -1 +0,0 @@
-toppar_*
```
j3mdamas: add it back with `./toppar_*` and that's all.

stefdoerr: done

PR #890, opened 2019-06-21
Parameterize: truncate decimals of diagnostic energy writing

stefdoerr:
htmd-data https://github.com/Acellera/htmd-data/pull/40

### DIFF
--- a/htmd/parameterization/cli.py
+++ b/htmd/parameterization/cli.py
@@ -512,19 +512,19 @@ def _printEnergies(molecule, parameters, filename):
     string = '''
 == Diagnostic Energies ==
 
-Bond     : {BOND_ENERGY:12.6g} kcal/mol
-Angle    : {ANGLE_ENERGY:12.6g} kcal/mol
-Dihedral : {DIHEDRAL_ENERGY:12.6g} kcal/mol
-Improper : {IMPROPER_ENERGY:12.6g} kcal/mol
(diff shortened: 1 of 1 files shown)


htmd/parameterization/cli.py
```diff
@@ -519,12 +519,12 @@ def _printEnergies(molecule, parameters, filename):
 Electro  : {ELEC_ENERGY:12.6g} kcal/mol
 VdW      : {VDW_ENERGY:12.6g} kcal/mol
 
-'''.format(BOND_ENERGY=energies['bond'],
-           ANGLE_ENERGY=energies['angle'],
-           DIHEDRAL_ENERGY=energies['dihedral'],
-           IMPROPER_ENERGY=energies['improper'],
-           ELEC_ENERGY=energies['elec'],
-           VDW_ENERGY=energies['vdw'])
+'''.format(BOND_ENERGY=np.round(energies['bond'], decimals=5),
```
raimis: It looks inconsistent, you are rounding to 5 decimal places, but the formatting is set to `12.6g`. Wouldn't setting it to `12.5f`, the rounding will be done implicitly.

stefdoerr: No, that would cause issues with large numbers not using scientific notation. I played around with it a bit

stefdoerr: Unless you don't care about such large energies. In which case sure I can switch from `g` to `f` formatting

PR #905, opened 2019-08-23
removed requests library and updated registration code

stefdoerr:
Need to wait for the registration website to work again to test it


stefdoerr:
Tested and it works

PR #910, opened 2019-08-29
Implement PlayQueue

raimis:
- [x] Implement PlayQueue
  - [x] Execution
  - [x] Logging

### DIFF
--- a/htmd/queues/playqueue.py
+++ b/htmd/queues/playqueue.py
@@ -64,6 +64,9 @@ class PlayQueue(SimQueue, ProtocolInterface):
 
     def inprogress(self):
 
+        # TODO updated to use playmolecule.Job.getChildren
+        #      see https://github.com/Acellera/htmd/pull/910
+
         counter = 0
         for jobID in self._dirs:
             job = self._getSession().getJob(id=jobID)
--- a/htmd/queues/playqueue.py
+++ b/htmd/queues/playqueue.py
@@ -1,0 +1,111 @@
+# (c) 2015-2019 Acellera Ltd http://www.acellera.com
+# All Rights Reserved
+# Distributed under HTMD Software License Agreement
+# No redistribution in whole or part
+#
+import logging
+import os
+import shutil
+import tempfile
+import zipfile
+
+from htmd.queues.simqueue import SimQueue
+from protocolinterface import ProtocolInterface
+from protocolinterface.validators import Number, String
+
+logger = logging.getLogger(__name__)
+
+class PlayQueue(SimQueue, ProtocolInterface):
+
+    def __init__(self):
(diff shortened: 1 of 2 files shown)


htmd/queues/playqueue.py
```diff
@@ -0,0 +1,124 @@
...
+            elif status in (4, 5): # Completed or errored
+                pass
+            else:
+                raise ValueError('Unknown job status')
+
+        return counter
+
+    def retrieve(self):
+
+        for jobID, dir_ in list(self._dirs.items()):
```
stefdoerr: no need to wrap in `list()`

raimis: Bare in mind, `self._dirs.items()` gives an iterator, not a list. So ` self._dirs.pop(jobID)` modifies iterable used by that iterator in the loop, which is not allowed.

stefdoerr: ah kk

raimis:
> I don't see this working under most realistic scenarios.

It is already work in *Parameterize* App: https://github.com/Acellera/playmolecule-parameterize/pull/18

> If you close the python session or it dies you lose the job IDs and you cannot get them back. Also this forbids it's usage in asynchronous stuff like adaptive if we ever want to.

The other queues have the same characteristic.

> I would do it with a local DB (sqlite probably since it's best integrated in python) which keeps track of submitted jobs and folders.

It is just a thin wrapper of *PM* API. Why do we need that complexity?

stefdoerr:
That's not exactly true. SLURM and LSF both poll the server for currently running jobs with a given project name so the "inprogress" method works without need to track the jobs locally. Since I don't know if PM keeps track of jobs on a server, if it does, then we should also poll the server for those, otherwise keep track locally of the job ids

How does it work if you restart the python job?

raimis:
> That's not exactly true. SLURM and LSF both poll the server for currently running jobs with a given project name so the "inprogress" method works without need to track the jobs locally. Since I don't know if PM keeps track of jobs on a server, if it does, then we should also poll the server for those, otherwise keep track locally of the job ids

It could be done with `playmolecule.Job.getChildren`, but I haven't managed to work. It needs some serious debugging at *PM* level. Yes, you are right, now it works more like `LocalQueue` than `SlurmQueue`.

> How does it work if you restart the python job?

At the moment, *Parameterize* can only recover completed jobs, running jobs will be lost with any queues.

stefdoerr:
okay if we want it to work from the PM side then I will try to fix that once I'm back full time. For the moment we can leave it as is, just maybe add a TODO in the inprogress that we should make it work from the PM side.

raimis:
Done! Any more comments?

PR #918, opened 2019-09-06
 Parameterize: enable progress updates for QM jobs

raimis:
- [x] Enable progress updates for QM jobs

~~Depend on #917~~


raimis:
My current inspiration is PM API, I cannot do in any other way... :)

PR #919, opened 2019-09-06
Fix #528

### LINKED ISSUE #528: Fix change of directory with PCMSolver
From here: https://github.com/Acellera/htmd/pull/409#discussion_r139696856

Psi4 team fixed it here: https://github.com/psi4/psi4/pull/818

We should fix ours as soon as their fix is on main

raimis:
- [x] Fix #528


raimis:
Yes, the tests cover this case!

PR #924, opened 2019-09-13
Set default ncpus to 1. Rename ncpus to njobs to follow nomenclature

stefdoerr:
We have had a multitude of issues both personally and our users with parallelism. On some machines it works on others it doesn't. Usually it needs huge amounts of RAM for many trajectories which most users don't have. Adria had issues with relative paths in adaptive. I have had issues with filtering on a 32GB RAM machine, etc. Changes in the default backend of joblib also caused issues.

Instead of doing support and losing users we will make the default 1 thread and the users can find the options for parallelism in the docs.

Also since we might start issuing more HTMD version warnings, I made a way to permanently disable specific version warnings.


stefdoerr:
It's not totally unrelated as this PR increases the warnings. So I thought I'll fix that as well. Thanks

PR #928, opened 2019-09-18
Parameterize: fix PM progress for NNP

raimis:
- [x] Fix PM progress for NNP


raimis:
This is a small PM-specific change, so I self approve it!

PR #1011, opened 2021-09-15
Incorrectly set actionspace in AdaptiveBandit

DanielWicz:
In the docstring the actionspace is set by default as  `actionspace='metric'` (line 206). In the code it is set to tica.


stefdoerr:
The code is more up to date than the docs. I'll update the docs instead. Thanks for pointing it out.

stefdoerr:
https://github.com/Acellera/htmd/commit/163b86f84c7d11395484b5fb5cf6bf485b7bf1c3

PR #1078, opened 2024-01-03
old updated for bandit

AdriaPerezCulubret:
These are some old changes I had on a local repository from some time ago. They are old, but some are a bit substantial to AdaptiveBandit's algorithm. Also includes an option to use macrostates to compute rewards.


stefdoerr (approved):
looks good, thanks!

PR #1091, opened 2024-06-07
Mps adaptive

AdriaPerezCulubret:
Added compatibility with latest jobqueues update to use nividia_mps with Slurm in Adaptive sampling.

### DIFF
--- a/htmd/adaptive/adaptive.py
+++ b/htmd/adaptive/adaptive.py
@@ -162,10 +162,7 @@ class AdaptiveBase(abc.ABC, ProtocolInterface):
                 self._init()
                 input_dirs = natsorted(glob(path.join(self.inputpath, "e1s*")))
                 if not self.dryrun:
-                    if self.mps > 0:
-                        self._submit_mps(input_dirs)
-                    else:
-                        self.app.submit(input_dirs)
+                    self.app.submit(input_dirs)
             else:
                 # Retrieving simulations
                 logger.info("Retrieving simulations.")
@@ -214,10 +211,7 @@ class AdaptiveBase(abc.ABC, ProtocolInterface):
                             path.join(self.inputpath, "e" + str(epoch + 1) + "s*")
                         )
                         try:
-                            if self.mps > 0:
-                                self._submit_mps(newsims)
-                            else:
-                                self.app.submit(natsorted(newsims))
+                            self.app.submit(natsorted(newsims))
                         except Exception:
                             # If submitting fails delete all simulation inputs to not confuse _getEpoch()
                             for ns in newsims:
@@ -250,8 +244,12 @@ class AdaptiveBase(abc.ABC, ProtocolInterface):
             if os.path.exists(lockfile):
                 os.remove(lockfile)
 
-    def _submit_mps(self, newsims):
+    def _submit(self, newsims):
         from jobqueues.slurmqueue import SlurmQueue
+
+        if self.mps == 0:
+            self.app.submit(newsims)
+            return
 
         if not isinstance(self.app, SlurmQueue):
             raise ValueError(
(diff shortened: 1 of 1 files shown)


htmd/adaptive/adaptive.py
```diff
@@ -154,7 +161,25 @@ def run(self):
...
+                    if self.mps > 0:
+                        from jobqueues.slurmqueue import SlurmQueue
+
+                        if not isinstance(self.app, SlurmQueue):
+                            raise ValueError(
+                                "Multiple processes per simulation only supported with SlurmQueue."
+                            )
+                        input_dirs = natsorted(glob(path.join(self.inputpath, "e1s*")))
+                        for i in range(0, len(input_dirs), self.mps):
+                            if len(input_dirs[i : i + self.mps]) > 1:
```
stefdoerr, line 173: I'm rather sure you will get an index error here if your simulations are not divisible by self.mps
Use `min(i+self.mps, len(input_dirs))` to get the max of the range

htmd/adaptive/adaptive.py
```diff
@@ -154,7 +161,25 @@ def run(self):
...
                 if not self.dryrun:
-                    self.app.submit(natsorted(glob(path.join(self.inputpath, "e1s*"))))
+                    if self.mps > 0:
+                        from jobqueues.slurmqueue import SlurmQueue
+
+                        if not isinstance(self.app, SlurmQueue):
+                            raise ValueError(
+                                "Multiple processes per simulation only supported with SlurmQueue."
+                            )
+                        input_dirs = natsorted(glob(path.join(self.inputpath, "e1s*")))
```
stefdoerr, line 171: move this out of the two if/else conditions. It's identical

htmd/adaptive/adaptive.py
stefdoerr:
```py
batch = input_dirs[i:min(i+self.mps, len(input_dirs))]
self.app.submit(batch, nvidia_mps=len(batch) > 1)
```
Then you skip that else-if as well

AdriaPerezCulubret:
Forgot there's two submits, so now I moved the mps part into a function, it's a bit cleaner now

htmd/adaptive/adaptive.py
```diff
@@ -203,7 +214,10 @@ def run(self):
                             path.join(self.inputpath, "e" + str(epoch + 1) + "s*")
                         )
                         try:
-                            self.app.submit(natsorted(newsims))
+                            if self.mps > 0:
+                                self._submit_mps(newsims)
+                            else:
+                                self.app.submit(natsorted(newsims))
```
stefdoerr, line 220: just move the if/else inside of the _submit_mps (rename it to just _submit). Cleans up the `run` function.

```
    def _submit(self, newsims):
        from jobqueues.slurmqueue import SlurmQueue
        
        if self.mps == 0:
            self.app.submit(natsorted(newsims))
            return

        if not isinstance(self.app, SlurmQueue):
            raise ValueError(
                "Multiple processes per simulation only supported with SlurmQueue."
            )
        for i in range(0, len(newsims), self.mps):
            batch = newsims[i : min(i + self.mps, len(newsims))]
            self.app.submit(batch, nvidia_mps=len(batch) > 1)
```

AdriaPerezCulubret: Done!

htmd/adaptive/adaptive.py
```diff
@@ -153,8 +160,9 @@ def run(self):
             if epoch == 0 and self.generatorspath:
                 logger.info("Epoch 0, generating first batch")
                 self._init()
+                input_dirs = natsorted(glob(path.join(self.inputpath, "e1s*")))
                 if not self.dryrun:
-                    self.app.submit(natsorted(glob(path.join(self.inputpath, "e1s*"))))
+                    self.app.submit(input_dirs)
```
stefdoerr, line 165: you forgot to use the _submit method in the two places though :D

stefdoerr (approved):

